PR #87, opened 2026-04-22
fix(xlsx): custom tableStyle dxfs, stacked combo charts, scheme colors

yukiyokotani:
## Summary

Three rendering fixes discovered on sample-9 (\"Gift budget and tracker\"):

- **Stacked combo charts**: when a bar/line combo chart declared the bar first with `grouping=\"stacked\"` and a line later with `grouping=\"standard\"`, the line's grouping overwrote the bar's, so bars rendered side-by-side instead of stacked. Lock the grouping once a non-line series sets it (ECMA-376 §21.2.2.17).
- **Chart series scheme colors**: series `<a:schemeClr val=\"accentN\">` was being dropped, so series fell back to palette defaults instead of the file's theme accents. Resolve schemeClr against the theme color table during parse.
- **Custom `<tableStyle>` elements** (ECMA-376 §18.8.40): parse `wholeTable` / `headerRow` dxf indices from `xl/styles.xml/tableStyles`, forward them through `TableInfo` → `TableCellStyle`, and overlay the resolved dxf fill, font color, and horizontal/vertical borders. Built-in style names (e.g. `TableStyleLight18`) keep the existing accent-based renderer unchanged.

Adds `Border.horizontal` and `Border.vertical` to carry the inner-rule edges emitted only by tableStyle dxfs.

Known follow-ups (deferred; touch `packages/core`):
- `\$` currency prefix on chart value-axis tick labels (custom `<c:valAx>` numFmt formatCode).
- Line-chart marker visibility flag is forwarded but the overlaid series line in the bar+line combo still lacks circle markers.

## Test plan
- [x] tsc --noEmit clean
- [x] wasm-pack rebuild
- [x] Storybook preview: `xlsxviewer/PrivateExamples/sample-9` → \"Gift budget and tracker\" renders stacked bars, correct accent colors, pink header row, visible row dividers

### DIFF
--- a/packages/xlsx/parser/src/lib.rs
+++ b/packages/xlsx/parser/src/lib.rs
@@ -95,6 +95,17 @@ pub struct TableInfo {
     /// theme accents (e.g. `TableStyleLight18` → accent3 of theme1.xml). Used
     /// by the renderer to draw banding, header background, and rules.
     pub accent_color: String,
+    /// Dxf index for the `wholeTable` element of a custom `<tableStyle>`
+    /// (ECMA-376 §18.8.40). When set, its border/fill apply to every cell
+    /// of the table as a base layer. Built-in style names use the renderer's
+    /// accent-based fallback, not this field.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub whole_table_dxf: Option<u32>,
+    /// Dxf index for the `headerRow` element of a custom `<tableStyle>`.
+    /// Provides the header background fill, font color/weight, and any
+    /// vertical separator borders shown between header cells.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub header_row_dxf: Option<u32>,
 }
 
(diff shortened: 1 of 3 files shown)


yukiyokotani:
Follow-up commit 3e293c6:

**Text operands in `cellIs` conditional formatting** (ECMA-376 §18.18.15). `cellIs` rules previously only matched numeric cells, so text rules like `equal "Birthdays"` silently skipped every row (e.g. G7–G11 on the "Gift budget and tracker" sheet). Now parses each `<formula>` as a quoted string literal or a number, and compares case-insensitively using the full operator vocabulary (equal / notEqual / containsText / notContains / beginsWith / endsWith / between / notBetween).

**Deferred to a separate core PR:** Excel auto-bumps the chart y-axis max past `ceil(dataMax/step)*step` when data lands too close to the top (sample-9: dataMax=150, Excel axMax=$200 with step=$20; we currently stop at $160). The step value itself is correct. Matching the padding heuristic needs multi-sample calibration to avoid regressing other charts, so it's out of scope here.

PR #465, opened 2026-06-16
fix(core): extend scatter value axis to Excel's nice maximum

yukiyokotani:
## Bug
On sample-30.xlsx the two scatter charts' value (Y) axes were cut off **flush at the data max** instead of at Excel's rounded maximum:
- chart3 (「標準偏差/平均周期」): rendered top ≈ 0.11 vs Excel **0.12** (data max 0.1129)
- chart2: rendered top **3.5** vs Excel **4** (data max 3.5)

## Root cause
`renderScatterChart` set `yMax = data max` for an auto value axis and never applied the "nice maximum" rounding. The other cartesian renderers (`renderBarChart` / `renderLineChart` / `renderAreaChart`) already call `niceAxisMax` (axis-scale.ts) — the scatter renderer was the only one that didn't, so its value axis sat flush against the top point.

`niceAxisMax` implements Microsoft's documented auto-scale ("Ymax + ~5% of the data range, rounded up to the next major unit"), so this is the existing, spec-consistent helper — not a new heuristic.

## Fix
Apply `niceAxisMin`/`niceAxisMax` to the scatter auto Y bounds (guarded so an explicit `<c:valAx><c:scaling><c:min/max>` still wins). The existing Y gridline loop recomputes the major unit on the rounded range, so labels stay round.

## Verification (Storybook + Playwright, sample-30.xlsx)
- chart3 Y-axis: `0, 0.02, 0.04, 0.06, 0.08, 0.1, 0.12` → **top 0.12** ✓
- chart2 Y-axis: `0, 1, 2, 3, 4` → **top 4** ✓
- core typecheck clean; axis-scale unit tests pass; no console errors.

## VRT note
This changes every scatter chart with an **auto** value axis (the Y max now extends to the nice maximum). Any demo scatter-chart reference image will shift and need maintainer-gated `UPDATE_REFS=1` regen after merge. Charts with explicit `<c:scaling>` are unaffected.

### DIFF
--- a/packages/core/src/chart/renderer.ts
+++ b/packages/core/src/chart/renderer.ts
@@ -1406,6 +1406,21 @@ function renderScatterChart(ctx: CanvasRenderingContext2D, chart: ChartModel, r:
   if (chart.valMin != null) yMin = chart.valMin;
   else if (yMin > 0) yMin = 0;
   if (chart.valMax != null) yMax = chart.valMax;
+  // Auto value (Y) axis: ONE major unit (the "nice" step) drives both the
+  // rounded maximum and the gridlines — identical to bar/line/area. niceAxisMax
+  // adds ~5% headroom above the data max and rounds up to that step, so the top
+  // point sits below the top gridline (data 3.5 → step 0.5, max 4 → 0,.5,…,4;
+  // 0.1129 → step 0.02, max 0.12). The step is taken from the DATA range and
+  // reused for the gridline loop below — recomputing it on the rounded range
+  // would wrongly coarsen it (niceStep(4)=1 vs the 0.5 Excel uses for 0–4).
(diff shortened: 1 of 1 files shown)


yukiyokotani:
**Updated:** use ONE major unit for the value axis (matching bar/line/area).

The first revision rounded the max to 4 but let the gridline loop recompute the step on the rounded range (`niceStep(4)=1`), giving `0,1,2,3,4`. Excel (and bar/line/area) use the step from the **data** range (`niceStep(3.5)=0.5`) for both the rounded bound and the gridlines. Now reused consistently → **chart2: `0,.5,…,4` (step 0.5)**, chart3: `0,.02,…,.12` unchanged. Verified on sample-30.

Note added in code: the auto major unit isn't specified by ECMA-376 (Excel-proprietary); `niceStep` approximates it and can differ from Excel by one step on some ranges.

PR #466, opened 2026-06-16
fix(xlsx): scale cell multi-line line-height by cell scale (zoom)

yukiyokotani:
## Bug
In a cell with multi-line text — e.g. the merged cell **I3:K7** in sample-30.xlsx — the line spacing does **not** follow the display scale. At zoom < 100% the glyphs shrink but the inter-line gaps stay fixed, so they grow *relatively* (the text spreads out). Excel scales the whole cell uniformly.

## Root cause
Glyphs are sized via `buildFont(font, cs)` → `font.size * PT_TO_PX * cs` (cs = cell scale). But every **font-size-derived vertical-positioning metric** was computed *without* `cs` — `font.size * PT_TO_PX * 1.2` (line height), the stacked-text char height, underline/strike Y offsets, and super/subscript baseline shifts. The cell box (`cy`/`cellH`, from the cs-scaled grid) *is* cs-scaled, so the metrics were mismatched at any zoom ≠ 100%.

## Fix
Multiply every such metric by `cs` — **13 sites** across both the merged-anchor pre-pass and the in-viewport text path (line heights, stacked `charH`, `rSizePx`/`sizePx` decoration offsets, super/sub `baseSizePx`/`baseSizePxOrig`). Width/`measureText` logic is untouched (already cs-scaled via the cs font).

## Why it's safe for VRT
`* cs` with `cs === 1` (the default `cellScale`, i.e. 100% zoom and what the VRT renders at) is a mathematical no-op (`round(x*1.2*1) === round(x*1.2)`). So all reference renders are **byte-identical** — only zoomed rendering changes. No reference regen needed.

## Verification (Storybook + Playwright, sample-30.xlsx at scale 0.5)
- **Before**: I3:K7 showed 3 lines with large gaps (spacing fixed, font shrunk).
- **After**: same cell fits 5+ lines with tight, proportional spacing; long right-side merged paragraphs pack correctly; data cells unchanged.
- xlsx typecheck clean.

### DIFF
--- a/packages/xlsx/src/renderer.ts
+++ b/packages/xlsx/src/renderer.ts
@@ -565,6 +565,15 @@ function blendHex(fgHex: string, bgHex: string, fgCoverage: number): string {
   return `rgb(${r},${g},${b})`;
 }
 
+/** Vertical pixel metric for cell text: point size → device px at the current
+ *  cell scale `cs`. `factor` is the line-height / char-height multiplier (1.2
+ *  for wrapped lines, 1.1 for stacked chars, 1.0 for decoration/baseline
+ *  offsets). Centralizes the `* cs` factor so a new vertical-metric draw site
+ *  can't silently omit it. (Glyph SIZE uses buildFont's floored variant.) */
+function vMetricPx(sizePt: number, cs: number, factor = 1): number {
+  return Math.round(sizePt * PT_TO_PX * factor * cs);
+}
+
 function buildFont(font: CellFont, cs = 1): string {
   const style = font.italic ? 'italic ' : '';
   const weight = font.bold ? 'bold ' : '';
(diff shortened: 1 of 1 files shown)


yukiyokotani:
Per the cross-package + duplication review: extracted `vMetricPx(sizePt, cs, factor=1)` and converted all 13 `Math.round(size * PT_TO_PX * factor * cs)` sites to it (commit d6903a8). Behavior-preserving (identical formula → byte-identical at every scale, no VRT change); making `cs` a required arg means a new vertical-metric draw site can't silently omit it — the exact bug class this PR fixes. Reviewers also confirmed pptx (bakes scale into sizePx) and docx (centralized lineBoxHeight(scale)) do **not** share this bug.

PR #469, opened 2026-06-16
fix(core): feather soft edges by blurring a separate silhouette layer

yukiyokotani:
`applySoftEdge` in `packages/core/src/shape/effects.ts` rendered the DrawingML soft-edge effect (ECMA-376 §20.1.8.31, `CT_SoftEdgesEffect` — the PowerPoint "Soft Edge Rectangle" picture preset) by blurring a *filled silhouette in place* while the shape's own clip was active. A clipped filtered fill produces no feather: the clip limits the blur kernel's sample region to the solid fill, so there is nothing transparent for the kernel to blur toward and the edge stays hard.

This change renders the solid silhouette onto a separate auxiliary layer first (unblurred, so the clip is harmless), then composites that layer through the blur filter via `drawImage`. Blurring a whole layer lets the kernel sample the transparent surround, producing a real edge falloff. A device-pixel-space fallback path (single in-place blur) is kept for environments where a second aux canvas is unavailable.

Verified visually against sample-11 slide 2's soft-edged photo versus the PowerPoint-exported PDF: the perimeter now feathers as expected instead of cutting off sharply. The existing `applySoftEdge` unit test (which encoded the old single-aux `destination-in` contract) is updated to assert the new two-layer pipeline, and `setTransform` is added to the recording-context test stub. `pnpm --filter @silurus/ooxml-core exec tsc --build` and `pnpm run lint` both pass; the full `@silurus/ooxml-core` vitest suite is green (208 tests).

### DIFF
--- a/packages/core/src/shape/effects.test.ts
+++ b/packages/core/src/shape/effects.test.ts
@@ -15,9 +15,25 @@ import type { Shadow, SoftEdge, Reflection } from '../types/common';
  */
 class RecordingCtx {
   ops: Array<{ op: string; args?: unknown[] }> = [];
-  filter = 'none';
-  globalCompositeOperation = 'source-over';
   fillStyle: unknown = '#000';
+  // Every composite-op / filter assignment is recorded, not just the last value,
+  // so a multi-pass pipeline (e.g. set destination-in + blur, then reset) can be
+  // asserted by history rather than the final field value.
+  usedCompositeOps: string[] = [];
+  usedFilters: string[] = [];
+
+  #filter = 'none';
+  #gco = 'source-over';
+
+  get filter() { return this.#filter; }
+  set filter(v: string) { this.#filter = v; this.usedFilters.push(v); }
+  get globalCompositeOperation() { return this.#gco; }
+  set globalCompositeOperation(v: string) { this.#gco = v; this.usedCompositeOps.push(v); }
+
+  /** Filter strings that were a blur(...) (i.e. excluding the 'none' resets). */
+  get usedBlurFilters(): string[] {
+    return this.usedFilters.filter(f => f.startsWith('blur('));
+  }
 
   save() { this.ops.push({ op: 'save' }); }
   restore() { this.ops.push({ op: 'restore' }); }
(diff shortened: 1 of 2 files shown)


yukiyokotani:
### Refinement: symmetric feather matching PowerPoint

The first commit on this branch masked the hard-clipped image with a blurred silhouette. That only feathers **inward** (the image has no pixels outside the rect), so the outer edge stayed a crisp hard step with just a thin white haze inside it — not PowerPoint's soft edge.

This commit reworks `applySoftEdge` (`packages/core/src/shape/effects.ts`) to feather **symmetrically** about the geometry, matching the rasterised falloff PowerPoint produces for `softEdge` (ECMA-376 §20.1.8.31 / §20.1.8.53):

1. Paint the sharp shape.
2. Build an **opaque colour layer** that extends past the rect via an edge-clamp stretch (`drawImage` of the bbox grown by `rad` on every side), then redraw the sharp image on top to keep the interior crisp. The layer is opaque across the whole feather band.
3. Replace its alpha with the **blurred silhouette** (`destination-in` + `blur(rad/3)`). Because the colour layer is fully opaque under the band, the result is exactly the blurred-silhouette alpha → a clean, symmetric Gaussian falloff, with no hard outer step and no doubly-faded halo.

**Blur std-dev = `rad/3`.** The soft-edge `rad` is the kernel extent (~3σ), and CSS `blur()` takes the std-dev, so σ = rad/3.

**Pixel-profile verification** against `sample-11`'s PowerPoint-exported PDF (edge alpha samples, outer → inner):

| | profile |
|---|---|
| Ours | `255 252 233 184 122 74 55` |
| PowerPoint | `255 253 239 200 143 89 58` |

Smooth monotonic ramp with no outer cliff, tracking PP within a few levels per step.

**Gates (local):** core `vitest` 208/208 green (incl. updated `applySoftEdge` test asserting the 3-aux-canvas / edge-clamp-stretch / `destination-in` + `blur` pipeline), `tsc --build` on `@silurus/ooxml-core` exit 0, root `ast-grep` lint exit 0. No reference images or media touched.

PR #474, opened 2026-06-16
fix(pptx): honor per-slide/layout clrMapOvr overrideClrMapping (§19.3.1.7)

yukiyokotani:
## Problem

PR #471 added per-slide theme/master resolution (`build_master_bundle` + `bake_clr_map`) but bakes only the slide **master**'s `<p:clrMap>` (ECMA-376 Part 1 §19.3.1.6). A slide or slide layout may carry `<p:clrMapOvr><a:overrideClrMapping .../></p:clrMapOvr>` (CT_ColorMappingOverride, §19.3.1.7; the override is a full CT_ColorMapping, §20.1.6.8) that **replaces** the master's logical-name→slot mapping for that slide/layout. `clrMapOvr` / `overrideClrMapping` / `masterClrMapping` were referenced nowhere in `packages/pptx`, so a slide using `overrideClrMapping` was silently resolved with the master's `clrMap` — #471's "honor clrMap" was not honored at the per-slide level.

## Fix

- Refactor `bake_clr_map` into reusable pieces with **byte-for-byte unchanged** behavior for existing callers:
  - `CLR_MAP_LOGICALS` — the 12 logical→default-slot pairs.
  - `parse_clr_map_node` — reads the 12 CT_ColorMapping attrs from any node (shared by `<p:clrMap>` and `<a:overrideClrMapping>`, identical attr set).
  - `apply_clr_map` — the baking loop; reads the raw scheme slots and writes the logical keys, leaving raw slots intact.
- Add `parse_clr_map_ovr`: `Some(map)` for `<a:overrideClrMapping>`; `None` for `<a:masterClrMapping/>` (§20.1.6.6) or an absent `<p:clrMapOvr>` (inherit the master).
- In `parse_presentation`, resolve the effective override with precedence **slide > layout > master** (the slide→layout→master mapping chain), re-bake it onto a clone of the master-baked theme, and thread it into `parse_slide` via a new `theme_override` argument. Re-baking is correct because the raw scheme slots (`dk1`/`lt1`/`dk2`/`lt2`/`accentN`/`hlink`/`folHlink`) are never overwritten, so the override resolves against the intact palette.
- Add a unit test mirroring the existing `clr_map_*` tests: a slide whose `clrMapOvr` swaps `bg1`/`tx1` resolves `schemeClr val="tx1"` to `lt1` (`#FAFAFA`), a different slot than the master's `dk1` (`#222222`).

## Test plan

All three CI gates run from the repo root and pass:

- `cargo fmt --all --check` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — no warnings
- `cargo test --workspace` — pptx_parser **64 passed; 0 failed** (+1 new), docx 43, ooxml-common 35, mcp-server 41, xlsx 50; 0 failures

### DIFF
--- a/packages/pptx/parser/src/lib.rs
+++ b/packages/pptx/parser/src/lib.rs
@@ -2009,42 +2009,52 @@ fn parse_theme_colors(xml: &str) -> HashMap<String, String> {
 /// default mapping is applied: bg1=lt1, tx1=dk1, bg2=lt2, tx2=dk2, accentN
 /// identity, hlink/folHlink identity. The raw slot keys (dk1, lt1, …) added by
 /// `parse_theme_colors` are left untouched, so canonical lookups still work.
-fn bake_clr_map(theme: &mut HashMap<String, String>, master_xml: Option<&str>) {
-    // logical name → default scheme slot (used when clrMap is missing the attr).
-    const LOGICALS: &[(&str, &str)] = &[
-        ("bg1", "lt1"),
-        ("tx1", "dk1"),
-        ("bg2", "lt2"),
-        ("tx2", "dk2"),
-        ("accent1", "accent1"),
-        ("accent2", "accent2"),
-        ("accent3", "accent3"),
-        ("accent4", "accent4"),
-        ("accent5", "accent5"),
-        ("accent6", "accent6"),
-        ("hlink", "hlink"),
-        ("folHlink", "folHlink"),
-    ];
+/// The 12 logical scheme names of a `CT_ColorMapping` (ECMA-376 §19.3.1.6),
+/// paired with the scheme slot each maps to by default when a `<p:clrMap>`
+/// attribute is absent: bg1=lt1, tx1=dk1, bg2=lt2, tx2=dk2, accentN identity,
+/// hlink/folHlink identity. Shared by `<p:clrMap>` (§19.3.1.6) and
+/// `<a:overrideClrMapping>` (§20.1.6.8), which carry the identical attribute set.
+const CLR_MAP_LOGICALS: &[(&str, &str)] = &[
+    ("bg1", "lt1"),
+    ("tx1", "dk1"),
+    ("bg2", "lt2"),
+    ("tx2", "dk2"),
+    ("accent1", "accent1"),
+    ("accent2", "accent2"),
+    ("accent3", "accent3"),
+    ("accent4", "accent4"),
+    ("accent5", "accent5"),
+    ("accent6", "accent6"),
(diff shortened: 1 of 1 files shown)


yukiyokotani:
## Independent review round → fixes (commit 2aaae06)

Ran a critical, independent multi-agent review of this change across four dimensions — ECMA-376 spec fidelity, single responsibility, duplication, and cross-package (docx/xlsx/pptx) consistency. Findings and resolutions:

**Fixed**
- **Master-inherited scheme colors now honor the override (spec gap).** The first commit applied the override only to the slide's own shapes; `build_master_bundle` freezes the theme-dependent master fields (`master_bg`, master txStyles colors, master bullet colors) against the master's own clrMap, so a slide's `clrMapOvr` did not flip a master-inherited background or placeholder color. Per §20.1.6.8 the override is used "in place of" the master mapping for the **whole** slide, so these are now recomputed against the slide's effective mapping (new `EffectiveMaster`, built in `parse_presentation` where `zip` is available since a master `<p:bg>` may reference a blip). This also removed the lone `theme_override` parameter on `parse_slide`.
- **Perf:** `parse_clr_map_ovr` now fast-rejects (substring guard) to avoid a second roxmltree parse of the slide XML on the common no-override path.

**Verified / deliberately kept**
- **`masterClrMapping` precedence.** A reviewer suggested an explicit `<a:masterClrMapping/>` on a slide should bypass the layout's override. Kept the inherit behavior: §20.1.6.6 / §19.3.1.7 say "use the master mapping", but Annex L.3.2.5 states a layout's color-map override "is inherited by all presentation slides that utilize this layout", and PowerPoint serializes `<a:masterClrMapping/>` on ordinary non-overriding slides — so bypass would break layout-override inheritance for the common case. Added a spec-cited comment and tests pinning it.
- **Cross-package consistency:** the clrMap/clrMapOvr machinery is PresentationML-specific and correctly pptx-local (`ooxml-common`'s `ColorResolver` doc explicitly keeps scheme resolution per-crate). The one real divergence — duplicated logical→slot default tables across pptx/docx/xlsx — is pre-existing tech debt; a future `ooxml-common` hoist, out of scope here.

**Tests:** +6 (pptx 64 → 70). Master-inherited background flips under an override (+ no-override control); a layout override is inherited via `masterClrMapping` and via an absent `clrMapOvr` (+ no-layout-override control); the `parse_clr_map_ovr` fast path preserves all results. All three CI gates green locally (`cargo fmt --all --check`, `clippy --workspace --all-targets -D warnings`, `test --workspace`).

PR #482, opened 2026-06-17
refactor(core): unify duplicated CJK break-range definitions

yukiyokotani:
## What

Unifies four hand-maintained copies of the CJK / ideographic Unicode break range into a single shared predicate in `@silurus/ooxml-core`.

The same range was duplicated across the renderers and had **drifted**:

| Site | Hangul upper bound |
|---|---|
| `packages/pptx/src/text-justify.ts` (`CJK_RE`) | U+D7FF |
| `packages/pptx/src/renderer.ts` (layout `CJK_RE`) | U+D7FF |
| `packages/docx/src/renderer.ts` (`hasCJKBreakOpportunity`) | U+D7AF |
| `packages/xlsx/src/renderer.ts` (`isCJKCodePoint`) | U+D7AF |

This violated the repo principle that statically-checkable rules belong in shared code, not in prose (the contract was enforced only by a "Mirrors the CJK ranges the layout uses" comment).

## How

- New `packages/core/src/text/cjk-ranges.ts` exports `isCjkBreakChar(cp: number): boolean` (beside the existing `kinsoku/` and `bidi/` text primitives), re-exported from `@silurus/ooxml-core`.
- Canonical ranges: **U+3000–9FFF, U+AC00–D7A3, U+F900–FAFF, U+FF00–FFEF**.
- **Hangul bound U+D7A3** = the exact Hangul Syllables block. U+D7A4–D7AF are unassigned and U+D7B0–D7FF is Hangul Jamo Extended-B (combining jamo); neither should be a standalone break/justify unit, so both the old D7FF and D7AF were over-broad. The three demo samples contain **zero** code points in U+D7A4–D7FF, so the narrowing is behaviour-neutral for the corpus.
- **U+3000** is included (wrap/break callers need it). It is harmless for the justifier: `text-justify.ts` classifies U+3000 as whitespace and consumes it as one inter-word gap *before* the CJK test is reached, so it is never double-counted. Two regression tests pin `justifyLine` output for lines containing U+3000.
- All four sites now call the shared predicate.

## Verification

- `vitest run` (core + pptx + docx + xlsx): **518 passed** (incl. 17 new core boundary tests + 2 new text-justify regression tests).
- `tsc --build`: clean. `pnpm run lint` (ast-grep): clean.
- VRT (`pnpm build:wasm && pnpm vrt`): all **demo-backed** tests green (pptx/docx/xlsx). `private/sample-*` refs are gitignored (absent in a fresh checkout) and out of scope.

## Follow-up

The parallel PR `fix/docx-distribute-cjk` adds `packages/docx/src/text-distribute.ts`, which currently carries its own local `isCJKCodePoint` (it could not import this core predicate because the file did not exist on `main` when it was written). The two PRs merge cleanly (different regions — verified with `git merge-tree`). **After both land, fold `text-distribute.ts`'s `isCJKCodePoint` into `core.isCjkBreakChar`** (~2-line change) so docx has a single CJK source.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -154,3 +154,4 @@ export {
   kinsokuAdjustedSplit,
   crossRunKinsokuRetract,
 } from './text/kinsoku';
+export { isCjkBreakChar } from './text/cjk-ranges';
--- a/packages/core/src/text/cjk-ranges.test.ts
+++ b/packages/core/src/text/cjk-ranges.test.ts
@@ -1,0 +1,32 @@
+import { describe, it, expect } from 'vitest';
+import { isCjkBreakChar } from './cjk-ranges.js';
+
+describe('isCjkBreakChar', () => {
+  // [code point, expected] — exercises every range edge (low−1, low, high,
+  // high+1) plus the unassigned/Jamo-Extended-B gap above Hangul Syllables.
+  const cases: [number, boolean][] = [
+    [0x2fff, false], // just below U+3000
+    [0x3000, true], // ideographic space (range start)
+    [0x3001, true], // 、 (former pptx-justify start)
+    [0x9fff, true], // CJK Unified Ideographs end
+    [0xa000, false], // just above U+9FFF
+    [0xac00, true], // Hangul Syllables start
+    [0xd7a3, true], // Hangul Syllables end (canonical upper bound)
+    [0xd7a4, false], // unassigned — was wrongly included by the D7AF/D7FF copies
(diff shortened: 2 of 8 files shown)


yukiyokotani:
Sibling: #483 (docx CJK distribute). Recommended merge order — land this first so `core.isCjkBreakChar` exists; then #483's local `isCJKCodePoint` in `text-distribute.ts` can be folded into it (the two merge cleanly per `git merge-tree`).

PR #483, opened 2026-06-17
fix(docx): justify CJK both/distribute via inter-character pitch (§17.18.44)

yukiyokotani:
## What

Implements spec-faithful justified alignment for **CJK** text in the docx renderer. `jc="both"` / `"distribute"` (ECMA-376/ISO-29500 §17.18.44 `ST_Jc`) previously distributed slack across **ASCII spaces only**, so a space-free CJK paragraph was never justified — Japanese 均等割り付け (distribute) did nothing.

## Spec

§17.18.44:
- `both`: "shall only affect the inter-word spacing ... and not the inter-character spacing within each word."
- `distribute`: "shall equally affect the inter-word spacing ... as well as the inter-character spacing ... additional character pitch shall be added to all characters."

The "inter-word only" clause for `both` is **Latin-scoped**. Ground truth: in the `private/sample-9` Word PDF, a wrapped CJK `both` line containing **zero ASCII spaces** is filled exactly to the right text margin (`pdftotext -bbox`: xMax = the page right margin; the natural last line stops short). The only mechanism that fills a space-free line is inter-CJK pitch — so **Word's `both` widens inter-word spaces AND inter-CJK boundaries**, differing from `distribute` only on the last line (both: natural, distribute: filled). This mirrors the pptx justify kernel (verified there against PowerPoint), now confirmed for WordprocessingML against Word.

## How

- New pure kernel `packages/docx/src/text-distribute.ts` → `distributeLineSlack(...)`: walks the line's code-point stream, opens gaps at inter-word whitespace (ASCII + U+3000) and inter-CJK boundaries, returns per-segment split/gap info + `perGap`. Preserved: leading 字下げ indent fixed, the final segment opens no gap (Σgaps == slack, final glyph lands on the margin), negative-slack compression opens spaces only.
- `renderer.ts`: the space-only `extraPerSpace` / `countTrailingSpaces` logic is replaced by `distributeLineSlack`; CJK segments draw in pieces advancing `perGap` between ideographs; highlight / underline / strike / ruby / `onTextRun` span the stretched width. Latin justify, ruby, track-changes, numbering and bidi paths are unchanged.
- 20 unit tests for the kernel.

## Verification

- `vitest run packages/docx`: **86 passed** (66 prior + 20 new).
- `tsc --build packages/docx`: clean.
- Headless skia before/after vs the actual Word-PDF ground truth:
  - `private/sample-9` p1 CJK `both` line: 514.7px → **523.0px (exact margin, matching the Word PDF's 523.0pt)**; last line stays natural.
  - Full-page pixel diff (before/after): demo/sample-1 (English), private/sample-7 (Arabic both), private/sample-8 (Arabic+LTR bidi) = **0.000% changed pixels** — Latin/Arabic/bidi byte-identical.
- Browser VRT not run (the VRT-covered demo sample has no CJK both/distribute content, and a worktree `@playwright/test` mismatch blocked the runner); the skia before/after vs the Word PDF gives equal-or-stronger coverage for this feature.

## Follow-up

`text-distribute.ts` carries a local `isCJKCodePoint` because the shared `core.isCjkBreakChar` (PR `refactor/cjk-range-dedup`) is not yet on `main`. After both merge, fold it into the core predicate (the two PRs merge cleanly — verified with `git merge-tree`).

### DIFF
--- a/packages/docx/src/renderer.ts
+++ b/packages/docx/src/renderer.ts
@@ -46,6 +46,10 @@ import {
   resolveFloatOverlap,
   skipPastTopAndBottom,
 } from './float-layout.js';
+import {
+  distributeLineSlack,
+  type SegStretch,
+} from './text-distribute.js';
 
 const HIGHLIGHT_COLORS: Record<string, string> = {
   yellow: '#FFFF00', cyan: '#00FFFF', green: '#00FF00', magenta: '#FF00FF',
@@ -2098,12 +2102,6 @@ function renderParagraph(
     para.alignment === 'distribute';
   const stretchLastLine = para.alignment === 'distribute';
 
-  const countTrailingSpaces = (s: string) => {
-    let c = 0;
-    for (let i = s.length - 1; i >= 0 && s[i] === ' '; i--) c++;
-    return c;
-  };
-
   // Bidirectional text. The paragraph's base direction comes from w:bidi
   // (ECMA-376 §17.3.1.6). We engage the (exact) bidi pass only when the base is
   // RTL or the line actually contains strong-RTL characters, so pure-LTR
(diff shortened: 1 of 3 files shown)


yukiyokotani:
Sibling: #482 (CJK range dedup → core). See the Follow-up section: `text-distribute.ts` keeps a local CJK predicate until #482's `core.isCjkBreakChar` lands on main.

yukiyokotani:
Updated after the second review round (rebased on #482, now merged):
- **Fix (regression):** RTL/bidi justified lines were skipped (`u.si >= lastDrawnSi` vs a visual-order `lastDrawnSi`) → now an exact match; LTR byte-identical. Added a kernel test pinning the bidi contract.
- **Dedup:** dropped the local `isCJKCodePoint` (6th copy) for the shared `core.isCjkBreakChar` from #482 — docx now has a single CJK source.
- **Deferred (documented in-code):** spec-literal `distribute` should also add inter-character pitch between Latin letters ("all characters on the line"); rare, needs a Word ground truth — tracked as follow-up, CJK unaffected.

PR #484, opened 2026-06-17
refactor(core): extract shared justified-line slack kernel

yukiyokotani:
## What

Extracts the justified-line slack distributor that pptx and docx had duplicated (~175 overlapping lines, ~75% of each kernel) into a single shared primitive in `@silurus/ooxml-core`. **Behavior-preserving — no pixel change.**

## Why

PR #481 (pptx `justifyLine`) and #483 (docx `distributeLineSlack`) independently implemented the same gap model: flatten the line to code points → content span → mark a gap at each inter-word space / inter-CJK boundary → `perGap = slack / total`. Two copies, two mirror-image test suites. With both on main, this is the inflection point to converge them.

## How

- **New `packages/core/src/text/line-distribute.ts`** → `distributeLineSlack(segments, slack, opts?)` (beside `kinsoku/` and `bidi/`), exported from `@silurus/ooxml-core`. Format-agnostic: returns the per-segment gap geometry (`{ splitBefore, trailingGap, internalStretch }` + `perGap`); no ECMA-376 policy lives here. Policy is injected via opts:
  - `isWhitespace` — PowerPoint counts every JS `\s`; Word counts only U+0020/U+3000. Each adapter injects its own.
  - `isGapChar` — `core.isCjkBreakChar` for expansion; `() => false` for docx's compression path.
  - `firstContentSi` / `lastDrawnSi` / `minPerGap` — leading-indent skip, the (exact-match, bidi-safe) final-segment exclusion, and the compression floor.
- **docx** `text-distribute.ts` → thin positional adapter over the core kernel; renderer untouched (same call, `SegStretch` re-exported from core). 247 → 106 lines.
- **pptx** `text-justify.ts` → thin adapter; `justifyLine` keeps its exact `(T & { jext })[]` signature and draw-piece output. It owns the DrawingML last-line policy + `slack <= 0.5` floor, injects PowerPoint's `\s` whitespace, and re-expresses the kernel's per-seg result as draw pieces. 169 → 144 lines.

### Subtlety worth noting (pinned by a test)

pptx and docx exclude the final segment differently. docx draws its visually-last segment separately and excludes it (`lastDrawnSi = visual.order[last]`, exact-match for bidi). pptx draws every segment in one loop, so it must let gaps open inside the last segment too — the final glyph is already suppressed by the kernel's content-span trim. The adapter passes `lastDrawnSi = segments.length` (a sentinel matching no real segment). A new pptx test pins the multi-glyph-final-segment case so this isn't "simplified" back to `length - 1` (which broke 10 tests).

## Verification (behavior-preserving)

- `vitest run`: **550 passed, 0 failed**. Justify subset: core 22 + pptx 16 + docx 8 = 46. The original 16 pptx + 19 docx kernel cases pass verbatim (the docx kernel cases moved to the core suite; docx keeps adapter-contract tests).
- `tsc --build` clean; `ast-grep scan` clean.
- **VRT: no pixel diffs** — docx demo 8/8, pptx demo 12/12 (private-sample 404s are gitignored absent files, not diffs).

## Scope

Only `packages/{core,pptx,docx}` touched; xlsx and `core/src/fonts/scripts.ts` (a separate font-selection predicate) untouched. Pre-existing follow-up carried over (out of scope): `distribute`'s pure-Latin intra-word pitch (§17.18.44 "all characters") remains unimplemented — CJK unaffected, needs a Word ground truth.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -155,3 +155,10 @@ export {
   crossRunKinsokuRetract,
 } from './text/kinsoku';
 export { isCjkBreakChar } from './text/cjk-ranges';
+export {
+  distributeLineSlack,
+  type DistributeSeg,
+  type DistributeResult,
+  type DistributeOptions,
+  type SegStretch,
+} from './text/line-distribute';
--- a/packages/core/src/text/line-distribute.test.ts
+++ b/packages/core/src/text/line-distribute.test.ts
@@ -1,0 +1,280 @@
+import { describe, it, expect } from 'vitest';
+import { distributeLineSlack } from './line-distribute.js';
+import type { DistributeSeg, DistributeResult } from './line-distribute.js';
+
+// distributeLineSlack is the shared "fill the line" kernel behind WordprocessingML
+// §17.18.44 (`both`/`distribute`) and DrawingML §20.1.10.59 (`just`/`dist`). Given
+// a line's segments (logical order) and its slack, it returns per-segment split
+// points + a trailing-gap flag and the per-gap px. Both gap families participate:
+// inter-word whitespace AND inter-CJK boundaries (either side a CJK / ideographic
+// glyph). The Word PDF for private/sample-9 pins the CJK case: a wrapped pure-CJK
+// `both` line with zero ASCII spaces still fills to the right margin (xMax 523.0pt
+// via pdftotext -bbox), only possible by widening inter-CJK boundaries.
+//
+// These tests exercise the kernel directly with its default (WordprocessingML)
+// predicates AND with the injected predicates the pptx adapter supplies.
+
+const seg = (text?: string): DistributeSeg => ({ text });
+
+/** Total px the kernel will add across the whole line = Σ over segments of
+ *  (internal gaps + trailing gap) × perGap. Must equal `slack` so the final
+ *  glyph lands on the right margin. */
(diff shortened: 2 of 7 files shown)


yukiyokotani:
Updated after independent critical review (2 agents: behavior-preservation + design).

**Design review: no blockers** — spec-first boundary exact, single-responsibility split clean, DRY genuinely eliminated (not relocated), bidi `===` invariant correct + tested, empirical `0.25` cap kept at the call site per repo policy.

**Behavior review found one delta** (not a regression, a strict fix): the pre-extraction `justifyLine` *dropped* a `text===''` segment — which in pptx is an inline OMML equation — from a justified line (never drawn, pen not advanced). The adapter now emits it (jext 0, `math` field preserved); gap distribution among real glyphs is unchanged. So this PR is **not a pure no-op**: it also fixes inline-equation-drop on justified pptx lines. Pinned with a regression test.

**Polish applied**: `JustifySeg = core.DistributeSeg` alias (one structural source); documented the `lastDrawnSi ≥ segments.length` sentinel as an explicit "exclude no segment" contract.

Verification after fixes: vitest **551 passed / 0 failed** (justify subset 47), tsc clean, ast-grep clean, VRT demo no pixel diffs.

Follow-up (out of scope): xlsx multi-line justify could become the 3rd consumer of this kernel once it grows a wrap engine.

PR #557, opened 2026-06-22
fix(pptx): render charts with absolute rel targets and multi-level category axes (#556)

### LINKED ISSUE #556: PptxViewer: chart renders as blank area while PDF conversion and sibling line chart render correctly (v0.64.3)
## Summary

A chart on a PowerPoint slide renders as a **blank area** in the canvas viewer, while other chart types on neighbouring slides of the same deck render correctly. The chart type involved is one listed as supported in the feature matrix (`bar, line, area, radar, waterfall`), so this looks like a rendering bug rather than an unimplemented feature.

## Version

- `@silurus/ooxml@0.64.3`
- Entry point: `@silurus/ooxml/pptx` → `PptxViewer`
- Environment: Chromium (browser), `devicePixelRatio` 2

## What renders vs. what doesn't

In the **same `.pptx` deck**:

- ✅ A **line chart** ("Revenue, gross profit & net profit by quarter") renders perfectly — axes, gridlines, legend, all three series.
- ❌ A chart on a **"Cumulative build-up"** slide renders as an **empty grey box** — the surrounding KPI cards and text render fine, only the chart area is blank.

Importantly, converting the **same file to PDF renders the missing chart correctly**, which isolates the problem to the canvas/WASM chart renderer for this specific chart type (the file and its chart data are intact).

## Expected behaviour

The chart on the "Cumulative build-up" slide should render in the canvas viewer, as it does in the PDF conversion and as the line chart on the sibling slide does.

## Actual behaviour

The chart area is blank; no chart is drawn. (No obvious console error observed — I will add console output and the exact chart XML below.)

## Reproduction

A sanitised `.pptx` (no real user data) reproducing the blank chart, plus before/after screenshots and the relevant `ppt/charts/chartN.xml` `<c:*Chart>` element, will be attached in a follow-up comment.

## Notes

- Bisected to: not yet — only confirmed present on `0.64.3` (a clean reinstall + cleared build cache still reproduces).
- Suspected to be a specific chart subtype (e.g. stacked `areaChart` or `area3DChart` / a combo chart). Exact `<c:*Chart>` element to follow with the repro file.

## supporting files

[Nimbus-FY2026-Business-Review.pptx](https://github.com/user-attachments/files/29213552/Nimbus-FY2026-Business-Review.pptx)

Image

Image

## interesting finding
- using libreoffice to convert to pdf give the graphs
- the macOS pptx previewer also hides the graphs

yukiyokotani:
Fixes #556.

## Problem

A chart on a PowerPoint slide rendered as a **blank grey box** in `PptxViewer`, while the PDF conversion and sibling charts rendered fine. The blank chart was the **area chart** ("Annual recurring revenue"); single-point bar/line charts in the same deck still showed something, which is why the reporter saw "line renders, area blank."

## Root causes (two independent spec gaps)

**1. Absolute (package-root) relationship Targets were not resolved.**
`resolve_path()` always joined the Target onto the source part's directory. An OPC Target beginning with `/` (e.g. `/ppt/charts/chart5.xml`, as this deck's generator emits) resolved to `ppt/slides/ppt/charts/chart5.xml` — absent from the archive — so `read_zip_str` failed and the chart graphicFrame was **silently dropped**. Per OPC / ECMA-376 Part 2 §9.3 a leading-`/` Target is a package-root-absolute part name and must ignore the base directory. `xlsx` (`resolve_zip_path`) and `docx` already handled this; pptx was the outlier. The fix also benefits absolute image/theme/layout Targets.

**2. Multi-level category axes truncated every series to one point.**
Category extraction only recognized `strCache/numCache/strLit/numLit`. A `<c:cat>` whose labels live in a `<c:multiLvlStrCache>` (ECMA-376 §21.2.2.95), where `<c:pt>` are nested under `<c:lvl>`, produced an empty `categories`, collapsing the shared point count to 1 (`categories.len().max(1)`) and dropping all but the first value of every series. A single-point **area** chart is a zero-width sliver → blank plot (the reported symptom). The fix reads the innermost level (first `<c:lvl>`).

## Verification

- New Rust unit tests (Red→Green): `resolve_path_handles_absolute_leading_slash_target`, `legacy_chart_parses_multi_level_category_axis`.
- `cargo fmt` / `clippy -D warnings` / `cargo test` (95 passed).
- Headless render of the repro deck (sample-14) compared against the PowerPoint PDF: area, line, and bar charts now render with full category labels and series. Before the fix, 0 of 9 charts parsed; after, all 9 parse with correct categories and full value series.

### DIFF
--- a/packages/pptx/parser/src/lib.rs
+++ b/packages/pptx/parser/src/lib.rs
@@ -1888,7 +1888,18 @@ fn find_rel_target_by_type(rels_xml: &str, type_suffix: &str) -> Option<String> 
 
 /// Resolve a relative path against a base directory inside the ZIP.
 fn resolve_path(base_dir: &str, target: &str) -> String {
-    let mut parts: Vec<&str> = base_dir.split('/').collect();
+    // A relationship Target beginning with "/" is a package-root-absolute part
+    // name (e.g. `/ppt/charts/chart5.xml`), not a reference relative to the
+    // source part's directory. Resolve it from the package root and ignore
+    // `base_dir`; otherwise the base is prepended, yielding a path that doesn't
+    // exist in the archive (OPC part names are root-anchored when they start
+    // with "/" — ECMA-376 Part 2 §9.3). xlsx::resolve_zip_path and docx's
+    // media path resolution already handle this; pptx was the outlier.
+    let mut parts: Vec<&str> = if target.starts_with('/') {
+        Vec::new()
+    } else {
+        base_dir.split('/').collect()
+    };
     for seg in target.split('/') {
         match seg {
             ".." => {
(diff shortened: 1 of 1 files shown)


yukiyokotani:
Added commit `8f3f6b2` — resolve absolute slide-part Targets from the package root via `resolve_path("ppt", …)`, the same OPC §9.3 fix class as the chart-rel resolution. This closes the latent `format!("ppt/{rel_target}")` gap flagged in review. New unit test `resolve_path_resolves_slide_targets_from_package_root`; cargo test / fmt / clippy green.

PR #584, opened 2026-06-27
fix(xlsx): reserve one line height for empty shape paragraphs

yukiyokotani:
## Problem

In `packages/xlsx/src/renderer.ts`, `drawShapeText`'s per-paragraph line builder starts `lineHeight` at `0` and only raises it from text/math runs (`lineHeight = Math.max(lineHeight, pxSize * 1.2)`). An **empty paragraph** — a `<a:p>` with no run and no `<a:br>` — therefore flushed a line of `height: 0`. A blank line inside a shape/textbox collapsed to nothing and pulled the following paragraphs up, misaligning text vertically.

This is the xlsx sibling of the docx empty-mark fix ([#582](https://github.com/yukiyokotani/office-open-xml-viewer/pull/582)) — surfaced via the repo's "1 bug = other formats may have the same hole" cross-package principle. ECMA-376 §21.1.2.2.6 / §21.1.2.3: the paragraph mark still reserves one line at its run size.

## Fix

In `flushLine`, when a line never had its height raised (`lineHeight === 0`, i.e. a blank line), seed it from the paragraph's effective size — the nearest text size in the paragraph (matching the inline-math `lastTextPt` inherit), else `DEFAULT_FONT_SIZE` — using xlsx's synthetic `1.2` leading / `0.85` ascent.

xlsx differs from docx: it has **no** real font metrics (no `fontBoundingBox` measurement, no win-metric correction table — that lives only in `packages/docx/src/font-metrics.ts`). So this is **not** "use real metrics"; it simply reserves a default single line, mirroring the text-run branch. pptx already handles the empty line correctly via its `maxSizePx === 0 → paraDefaultFontSizePx` fallback, so no pptx change is needed.

## Verification

The xlsx VRT corpus **cannot** exercise this path: a full scan of all 30 samples shows **no textbox mixes a blank line with following text** — every empty shape paragraph lives in an *entirely empty* textbox that draws no glyphs either way, so before/after pixels are identical. (Same kind of coverage gap as the documented docx-VRT-no-list gap.) The requested Playwright VRT also can't run in this environment — the config uses the `chrome` channel and Google Chrome isn't installed.

Instead, this PR adds a committed **headless skia probe** (`packages/node/src/xlsx-shape-empty-para.probe.test.ts`, same pattern as `xlsx-border-crisp.probe.test.ts`, skia-gated so CI skips it). It renders `[text, empty, text]` vs a `[text, text]` control through the **real** `drawShapeText` and asserts the gap between the two text lines grows by exactly one blank-line height:

| | control gap | test gap | delta | expected |
|---|---|---|---|---|
| **pre-fix** | 17.9px | 17.9px | **0.00** ❌ | 17.6px |
| **post-fix** | 17.9px | 35.9px | **18.00** ✅ | 17.6px |

"Above" sits at the same top in both runs, so the delta isolates exactly the empty paragraph's reserved height (font-metric offsets cancel). A genuine Red→Green regression guard. `drawShapeText` is exported to enable it.

- `pnpm --filter @silurus/ooxml-xlsx typecheck` — clean
- `npx vitest run packages/node/src` — probe + both border probes green (one unrelated `source-buffer-image` failure is a worktree-env artifact: missing skia `loadImage` postinstall glue; that test imports pptx and never touches the xlsx renderer)
- `ast-grep scan` — clean

### DIFF
--- a/packages/node/src/xlsx-shape-empty-para.probe.test.ts
+++ b/packages/node/src/xlsx-shape-empty-para.probe.test.ts
@@ -1,0 +1,162 @@
+/**
+ * Headless probe for the xlsx shape-text EMPTY-paragraph line-height fix.
+ *
+ * Background: in `renderer.ts` `drawShapeText`, the per-paragraph line builder
+ * starts `lineHeight = 0` and only raises it from text/math runs
+ * (`lineHeight = Math.max(lineHeight, pxSize * 1.2)`). An EMPTY paragraph (a
+ * `<a:p>` with no run and no `<a:br>`) therefore flushed a line of `height: 0`,
+ * so a blank line inside a textbox collapsed to nothing and pulled the
+ * following paragraphs up. ECMA-376 §21.1.2.2.6 / §21.1.2.3 — the paragraph
+ * mark still reserves one line at its run size. The fix seeds a blank line's
+ * height from the paragraph's effective size (nearest text size → the
+ * 11pt DEFAULT_FONT_SIZE) using xlsx's synthetic `1.2` leading (xlsx has no
+ * real font metrics, unlike docx PR #582).
+ *
+ * Why this probe and not the VRT: a full scan of the xlsx sample corpus shows
+ * NO textbox mixes a blank line with following text — every empty shape
+ * paragraph lives in an entirely-empty textbox that draws no glyphs either way,
+ * so the VRT cannot observe this fix (a coverage gap, like docx's list gap).
+ * This probe injects the missing case directly.
+ *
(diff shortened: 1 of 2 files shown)


yukiyokotani:
Closing as a duplicate of #583, which landed the byte-identical fix (same `flushLine` zero-height guard + `export drawShapeText`) from a parallel session minutes earlier. #583 also ships a CI-runnable mock-context regression test (`empty-paragraph-shape-text-height.test.ts`) that supersedes this PR's skia-gated headless probe. No unique content here worth keeping. The fix is already in main via #583.

PR #585, opened 2026-06-27
fix(xlsx): reserve one line height for blank lines in rich-text cells

yukiyokotani:
## Problem

In `packages/xlsx/src/renderer.ts`, `layoutRichTextLines` (the wrapped **cell** rich-text layout) flushes a line region at every hard break (LF), but its `flush` helper early-returned on an empty region:

```ts
const flush = () => {
  if (cur.length === 0) return;   // empty region dropped — never pushed
  ...
};
```

So a `wrapText` cell whose value contains consecutive newlines (`A\n\nB`, i.e. Alt+Enter twice), or a leading/trailing break, **dropped the blank line** and pulled every line below it up by one line height — vertical misalignment of wrapped cell text.

This is the cell rich-text analog of the shape-text empty-paragraph fix in #583 (and docx #582). The repo's [cross-package integrative principle](../blob/main/CLAUDE.md) flags one such bug as a signal to check the sibling surfaces — this is the third instance of the same shared concept.

## Ground truth (spec-faithful, no heuristic)

- **ECMA-376 §18.8.1** (`CT_CellAlignment` `@wrapText`): the text in a cell is line-wrapped within the cell.
- Cell rich text is stored as `<r>`/`<t>` runs (§18.4.7 / §18.4.12) with `xml:space="preserve"`, so an Alt+Enter break is a literal LF inside the run text.
- Each rendered line — **including a blank one** — reserves one single-line height, as tall as a one-character line of the same font. This is universal text-layout behaviour and exactly what the **plain single-run cell path already does** (`wrapTextLines` splits on `'\n'`, and an empty paragraph yields a `['']` line). The bug was unique to the rich-run path; this change aligns it.

## Fix

`flush` keeps its old behaviour for the soft-wrap (kinsoku) path (a line carried wholly to the next line must not fabricate a blank). A new `flushRegion(emptySizePt)` — used for hard breaks and end-of-value — emits a blank `RichLine` for an empty region, sized from the current run's effective font size (falling back to the cell default, then `DEFAULT_FONT_SIZE`), mirroring the shape-text seed in #583. `RichLine.maxFontSize` feeds `vMetricPx(size, cs, 1.2)` downstream, so a blank line now reserves exactly one line height.

Leading and trailing blank lines are preserved (matching Excel and the shape path); a value with no content and no breaks still produces no line.

## Tests / verification

- New CI-runnable mock-context test `packages/xlsx/src/empty-line-cell-rich-text-height.test.ts` (mirrors #583's `empty-paragraph-shape-text-height.test.ts`): asserts `A\n\nB` emits a middle blank line of the same height as a text line, that the blank line is sized from the effective run size (22pt → 22pt, not 0/default), and that leading/trailing blanks are kept while a truly empty value produces no line. **Red** before the fix (blank lines dropped), **green** after.
- Full xlsx unit suite green (147 tests).
- Root/xlsx typecheck clean.
- xlsx VRT: **67 passed / 0 failed** across all 30 private samples + demo. No reference images updated. (No sample contains a rich-run cell with a blank line, so the VRT can't exercise this path — same coverage gap as #583; the unit test is the behavioural guard.)

### DIFF
--- a/packages/xlsx/src/empty-line-cell-rich-text-height.test.ts
+++ b/packages/xlsx/src/empty-line-cell-rich-text-height.test.ts
@@ -3,7 +3,7 @@ import { layoutRichTextLines } from './renderer.js';
 import type { CellFont, Run } from './types.js';
 
 // ECMA-376 §18.8.1 (CT_CellAlignment) @wrapText: text in a cell is line-wrapped
-// within the cell. Cell rich text is stored as <r>/<t> runs (§18.4.7 / §18.4.12)
+// within the cell. Cell rich text is stored as <r>/<t> runs (§18.4.4 / §18.4.12)
 // with xml:space="preserve", so a hard line break authored with Alt+Enter is a
 // literal LF (U+000A) inside the run text. Each rendered line — INCLUDING a blank
 // line produced by consecutive breaks, or a leading/trailing break — reserves one
@@ -51,7 +51,7 @@ function layout(text: string, font?: Run['font']) {
   return layoutRichTextLines(makeCtx(), [{ text, font }] as Run[], BASE, 1, 100000);
 }
 
-describe('empty line height in cell rich text (§18.8.1 wrapText / §18.4.7 r)', () => {
+describe('empty line height in cell rich text (§18.8.1 wrapText / §18.4.4 r)', () => {
   it('a blank line between two text lines reserves one single-line height', () => {
     // Reference: "A\nB" — two single-line regions, no blank between.
     const ctrl = layout('A\nB');
--- a/packages/xlsx/src/empty-line-cell-rich-text-height.test.ts
+++ b/packages/xlsx/src/empty-line-cell-rich-text-height.test.ts
@@ -1,0 +1,108 @@
+import { describe, it, expect } from 'vitest';
+import { layoutRichTextLines } from './renderer.js';
+import type { CellFont, Run } from './types.js';
+
+// ECMA-376 §18.8.1 (CT_CellAlignment) @wrapText: text in a cell is line-wrapped
+// within the cell. Cell rich text is stored as <r>/<t> runs (§18.4.7 / §18.4.12)
+// with xml:space="preserve", so a hard line break authored with Alt+Enter is a
+// literal LF (U+000A) inside the run text. Each rendered line — INCLUDING a blank
+// line produced by consecutive breaks, or a leading/trailing break — reserves one
+// single-line height, exactly as Excel renders it and as the shape-text path now
+// does (PR #583, the xlsx sibling of docx PR #582).
+//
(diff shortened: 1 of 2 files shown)


yukiyokotani:
## Independent multi-dimension review (spec-first / single-responsibility / dedup / cross-package)

Reviewed by three independent agents across the requested dimensions. Findings adjudicated and applied in `174d271`:

**Applied (this PR):**
- **Spec citation fix** — the run element is ECMA-376 **§18.4.4 (`r`)**, not §18.4.7 (which is `rPr`). Corrected the test header/describe and the function docstring (which also wrongly cited the WordprocessingML §18.3.1.53 `w:r`); added §18.4.8 (`si`) / §18.4.9 (`sst`). §18.4.12 (`t`) / §18.8.1 (`wrapText`) were already correct (verified against the local ECMA-376 Part 1 PDF).
- **Cross-package consistency** — a blank line is now seeded from the **nearest preceding text run** size (`lastTextPt`, updated in `push`), genuinely matching the shape-text path's `lastTextPt` seed (`drawShapeText`, #583) rather than the run that happens to own the break token. Identical for the common single-run case; the "Mirrors `lastTextPt`" comment is now accurate.
- **Documented the empty-value contract** — `layoutRichTextLines('')` returns `[]` (no fabricated line), deliberately unlike the plain-text `wrapTextLines` (`split('\n')` → `['']`), since an empty cell has no glyphs to anchor.
- Trimmed the over-long flush/flushRegion comments to match the function's terse style.

**Confirmed sound (no change needed):**
- The `flush` / `flushRegion` split is justified — the kinsoku soft-wrap carry-down relies on `flush`'s empty early-return, so it must not emit a blank.
- No forbidden heuristic (the `×1.2` lives in the pre-existing `vMetricPx` and applies to all lines; a blank line just uses the same factor as a text line).
- **Should-extract-to-core: NO.** The docx/pptx/xlsx line-builders use deliberately different fidelity models (docx measures real `fontBoundingBox`; pptx/xlsx use analytic `size×1.2`) and are tightly coupled to each renderer's segment/kinsoku/bidi/math machinery — forcing a shared layout helper would be the false-abstraction the repo principle warns against. The genuinely-pure bits (`kinsokuAdjustedSplit`, `isCjkBreakChar`, `classifyFontGeneric`, `PT_TO_PX`) are already shared from core.

**Out of scope → tracked follow-up:** the **non-wrap** rich-text cell path (`else if (hasRichText)`, ~line 2140) lays all runs on one line and ignores `\n` entirely, unlike the plain non-wrap path which multi-lines on Alt+Enter. Real bug, but a distinct concern (hard-break multi-line, not blank-line height); fixing it here would break this PR's single responsibility, so it's filed separately.

Verification after the fixes: new test 5/5, full xlsx suite 146/146, xlsx typecheck clean, VRT 67/0 (no reference images updated).

PR #589, opened 2026-06-27
fix(xlsx): resolve wrap rich-text base direction per LF paragraph

yukiyokotani:
## Problem

`drawWrappedRichText` (the WRAP rich-text draw helper) computed the bidi base direction **once for the whole cell**:

```ts
const baseRtl = needBidi && cellBaseRtl(opts.readingOrder, runs.map((r) => r.text).join(''));
```

For soft-wrapped display lines that is correct — UAX#9: a soft wrap does **not** start a new bidi paragraph, so all display lines of one paragraph share a base direction.

But a cell can be **both** `wrapText` **and** contain hard breaks (Alt+Enter `LF`). `LF` (U+000A) is Unicode bidi class **B (paragraph separator)**, so each LF-delimited region is a separate bidi paragraph. Under `readingOrder=0` (Context / first-strong, ECMA-376 §18.8.1) each paragraph should resolve its **own** base direction from its **own** first strong character (UAX#9 P1–P3). The single cell-wide first-strong direction meant a wrapped cell whose paragraph 1 starts LTR and paragraph 2 starts with a strong-RTL char (Hebrew/Arabic) rendered **paragraph 2 under the wrong base direction**.

The non-wrap hard-break path (`drawMultiLineRichText` → `drawRichLine`) already resolves base direction per LF line; this aligns the wrap path with it.

Pre-existing; surfaced during independent review of #587 (which refactored the inline wrap loop into `drawWrappedRichText` without changing the computation).

## Fix

- `layoutRichTextLines` now tags each `RichLine` with its **LF-paragraph index** (`para`), which advances only at a hard break — soft-wrapped continuation lines keep their paragraph's index. All existing layout / blank-line / kinsoku / height behaviour is untouched.
- `drawWrappedRichText` resolves the base direction **per LF paragraph** from that paragraph's own text and indexes it by `line.para`. Explicit LTR/RTL (`readingOrder` 1/2) stays cell-wide; only Context (0/absent) varies between paragraphs. The pure-LTR gate is preserved per paragraph.

## Tests

- New `wrap-rich-text-context-bidi.test.ts` (mock-context unit test — the real guard, since this niche mixed-direction multi-paragraph case is not exercised by VRT):
  - paragraph 2 (RTL-first) is ordered under its **own** RTL base, not paragraph 1's LTR base (the bug, with a precondition asserting the two orders actually differ);
  - a single-paragraph mixed cell still uses its first-strong base (unchanged);
  - a blank line between paragraphs is still preserved in the drawn vertical advance.
- Full xlsx unit suite: **162 passed**.
- xlsx VRT: **67 passed, 0 failed** (1 pre-existing skip), no regressions.

### DIFF
--- a/packages/xlsx/src/bidi-line.ts
+++ b/packages/xlsx/src/bidi-line.ts
@@ -44,6 +44,23 @@ export function segmentsHaveRtl(segments: readonly unknown[]): boolean {
   return false;
 }
 
+/**
+ * Resolve whether one cell line / paragraph needs the bidi pass and its base
+ * direction, from the xf @readingOrder (§18.8.1) and the line/paragraph text.
+ * Gated so pure-LTR text (no strong-RTL char and not explicitly RTL) keeps the
+ * exact pre-bidi path: `needBidi` false ⇒ no UAX#9 reorder. `baseRtl` is only
+ * resolved when the pass will run. The single source of truth for the gate +
+ * base-direction rule shared by the non-wrap ({@link cellBaseRtl} per LF line)
+ * and wrap (per LF paragraph) rich-text paths — keeping the two in lockstep.
+ */
+export function resolveCellBidi(
+  readingOrder: number | undefined,
+  text: string,
+): { needBidi: boolean; baseRtl: boolean } {
+  const needBidi = readingOrder === 2 || RTL_GATE.test(text);
+  return { needBidi, baseRtl: needBidi && cellBaseRtl(readingOrder, text) };
+}
+
 export interface LineVisualOrder {
   /** Logical segment indices in visual (left-to-right) order. */
   order: number[];
--- a/packages/xlsx/src/renderer.ts
+++ b/packages/xlsx/src/renderer.ts
@@ -773,6 +773,12 @@ interface RichSeg {
 interface RichLine {
   segments: RichSeg[];
   maxFontSize: number; // pt (line-height source)
+  /** 0-based index of the LF-delimited paragraph (hard-break region) this line
+   *  belongs to. Soft-wrapped continuation lines share their paragraph's index;
+   *  it advances only at a hard break. Indexes the per-paragraph bidi base
+   *  direction the wrap draw path resolves — UAX#9: a soft wrap does NOT start a
+   *  new paragraph, but a hard break does. */
+  para: number;
 }
 
 /**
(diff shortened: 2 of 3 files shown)


yukiyokotani:
## Independent multi-dimension review → one change applied

Reviewed by four independent agents (spec-first / single-responsibility / duplicate-logic / cross-package). Summary:

- **Spec-first (ECMA-376 + UAX#9):** ✅ faithful. Confirmed `LF` (U+000A) is Unicode bidi class **B**, so UAX#9 P1 mandates each LF region is its own bidi paragraph; the core engine resolves one base direction per input string, so P1 segmentation correctly belongs in the renderer. §18.8.1 readingOrder reading verbatim-correct; all cited sections (§18.8.1 / §18.4.4 / §18.4.12 / §18.4.14) accurate; no heuristics.
- **Single responsibility:** ✅ tagging `RichLine.para` is the right seam — the LF-paragraph boundary is information only `layoutRichTextLines` holds post-tokenization; the alternative (re-splitting + re-tokenizing in the draw path) would couple worse.
- **Cross-package (docx/pptx):** ✅ the bug is genuinely **xlsx-specific**, with traced evidence: docx derives base direction from `<w:bidi>` (`para.bidi === true`, §17.3.1.6) and pptx from `<a:pPr rtl>` (`para.rtl === true`); `<w:br>`/`<a:br>` are intra-paragraph line breaks (not bidi-B separators), so one base direction per paragraph is spec-correct there. No latent instance, no `core` extraction (would be false abstraction — only xlsx needs LF-aware per-paragraph resolution).
- **Duplicate-logic (the one actionable finding, rated *Important*):** the `needBidi`/`baseRtl` gate was triplicated (`drawRichLine` + the new `drawWrappedRichText` block, the latter via an awkward `segmentsHaveRtl([{text:t}])` adapter). The short-circuit invariant (`needBidi && cellBaseRtl(...)`) must stay in lockstep between the wrap and non-wrap paths — the exact desync class this PR fixes.

**Applied** (commit `f7521f7`): extracted a pure `resolveCellBidi(readingOrder, text): { needBidi, baseRtl }` into `bidi-line.ts` (next to `cellBaseRtl`/`segmentsHaveRtl`) and wired both rich-text paths to it — single source of truth, retires the `[{text:t}]` adapter. Pure predicate, xlsx-local (not `core`) — confirmed not a false abstraction. Behaviour-preserving: xlsx unit **162/162**, VRT **67/0** with byte-identical sheet diffs, typecheck clean.

Declined (low-value / churn, reviewers concurred): CR/CRLF comment (not a bug — both paths treat `\r` identically; Excel persists Alt+Enter as bare LF), `paraTexts` rename, doc-comment dedup, and stamping bidi state onto the layout pass (would widen `layoutRichTextLines`'s contract — the coupling the SRP review cautioned against).

PR #598, opened 2026-06-28
fix(docx): collapse zero-before continuous-section spacer to match Word

yukiyokotani:
## Problem

In **sample-12** the heading **\"1. INTRODUCTION\"** rendered ~24pt (one blank line + ~6pt) too low compared with Word's PDF.

Root cause: at a **continuous section break** formed by empty paragraphs, our renderer drew a mark line box for the empty paragraph that carries the `sectPr`, **and** kept the preceding paragraph's `spaceAfter`. Word does neither when that section-mark paragraph has no `space-before` of its own — it collapses the mark to zero height and sits the empty run flush below the preceding content.

This is the same Word-interop family as the existing section-break-spacer suppression (#595/#597); that fix only dropped the spacer's *own* `before`, which is `0` in sample-12, so it had no effect there.

## Fix

Reconstructed clean-room from Word's measured output (no external implementation referenced):

- **`isCollapsedContinuousSpacer`** — a continuous-section spacer whose own `space-before` is `0` renders **no mark line box** (the section mark collapses). A spacer *with* a `before` keeps its box (the before manifests as the blank line).
- **`leadsCollapsedRun`** — the empty paragraph beginning the run sits **flush** below the preceding paragraph (the previous `spaceAfter` is dropped), so the transition collapses upward.
- The paginator and the paint pass skip the collapsed spacer **in lockstep** (page fill == paint), so pagination is unchanged.

Gated on `before == 0`, which is exactly what distinguishes the two journal templates.

## Verification (browser, real fonts — what the user sees)

Measured INTRODUCTION's Y against the Word PDF (`pdftotext -bbox`), anchored on the \"Keywords\" line:

| sample | before | after | Word | result |
|---|---|---|---|---|
| **sample-12** (spacer before=0) | 470.4pt | **445.8pt** | 446.2pt | ✅ fixed (one blank line, heading flush) |
| **sample-13** (spacer before=22) | 377.6pt | **377.6pt** | 378.1pt | ✅ unchanged (two blank lines kept) |

- `packages/docx` unit tests: **294 pass** (2 new in `section-break-spacer.test.ts` isolating each facet)
- Pagination unchanged: sample-12 → 3 pages, sample-13 → 5, sample-5 → 7
- `pnpm typecheck` (root) ✅ · docx VRT **21/21** ✅ (no demo/private regression; sample-5's continuous breaks unaffected)

## Spec note

This mimics **Microsoft Word's actual layout**, not an ECMA-376 clause: §17.3.1.29 mandates that every paragraph produces one mark line box and no clause documents a section-mark collapse. The model and its `before == 0` gate are pinned against the Word PDFs of both samples. Scope is DOCX-only — `continuous` section breaks / paragraph-level `sectPr` (§17.6) have no analogue in pptx/xlsx, so no cross-package change is warranted.

### DIFF
--- a/packages/docx/src/renderer.ts
+++ b/packages/docx/src/renderer.ts
@@ -1132,6 +1132,30 @@ export function computePages(
   const isContinuousSectionSpacer = (idx: number): boolean =>
     isSectionBreakSpacerAt(body, idx) && sectionKindFrom(idx + 2) === 'continuous';
 
+  // A continuous-section spacer whose OWN space-before is zero. Word renders NO
+  // paragraph-mark line box for it: the section mark collapses to zero height
+  // instead of occupying a blank line. (A spacer that DOES carry a space-before
+  // keeps its box — the before manifests as the blank line.)
+  //
+  // NOTE — this matches Microsoft WORD's observed layout, NOT a spec rule.
+  // §17.3.1.29 mandates that every paragraph produces one mark line box; no
+  // ECMA-376 clause (nor [MS-DOC] / [MS-OI29500]) documents a section-mark
+  // collapse. The model is reconstructed clean-room from Word's OWN output:
+  //   - sample-12's spacer is Normal with before=0. Word shows exactly ONE blank
+  //     line between "[Format…single line spacing]" and "1. INTRODUCTION" (the
+  //     user verified this by cursor-walk) and paints the heading ~24pt higher
+  //     than our prior two-blank-line layout — i.e. the spacer's own mark line is
+  //     absent (pdftotext -bbox: INTRODUCTION at 446pt, not 470pt).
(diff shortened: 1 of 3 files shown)


yukiyokotani:
Reverting: the `before===0` collapse gate over-fires. sample-13 has multiple OTHER zero-before continuous-section spacers (doc para [41] before the "MSR - Instructions…" title, plus [65]/[67]/[145]/[218]); collapsing them removed page-1 height and pulled the page-2 title onto page 1, overflowing into the footnote area (text collision). sample-12's spacer and sample-13's [41] are structurally identical (empty + continuous sectPr + before=0), so the trigger can't be pinned from local structure — this needs a cleaner (or spec-grounded) distinguisher before re-attempting. Branch reset; the sample-12 INTRODUCTION-position issue is parked, not the fragile heuristic.

PR #599, opened 2026-06-28
fix(docx): collapse zero-before continuous spacer + reserve a tall first-page footer (sample-12/13)

yukiyokotani:
## Summary

Two stacked fixes for sample-12 / sample-13, where the "1. INTRODUCTION" heading rendered ~24pt too low (sample-12) and the page-2 title overprinted the first-page footer (sample-13).

PR #598 carried **only** the first fix and was closed because that fix alone regressed sample-13: the vertical space it frees on page 1 pulled the page-2 title up onto page 1, where it overlapped the masthead footer. The real cause of the overlap is a separate pagination gap (a footer taller than its margin was never reserved). This PR ships **both** fixes together.

## 1. Collapse a zero-before continuous-section spacer (`a653dc8`)

An empty paragraph that carries a continuous section mark and has **no** space-before of its own renders **no** paragraph-mark line box: Word shows one blank line there, not two, and the following heading sits ~24pt higher. A spacer that *does* carry a space-before keeps its box (sample-13 is unaffected by this rule). The empty run that leads into the collapsed spacer also drops the preceding paragraph's space-after so the heading lands where Word places it.

- `isCollapsedContinuousSpacer` / `leadsCollapsedRun` in `renderer.ts`, `collapsedSpacer?: boolean` in `types.ts`, applied in lockstep across the paginate and paint passes.
- Reconstructed clean-room from Word's actual output (no spec section governs this interop behaviour; it is pinned by measurement).
- 2 unit tests in `section-break-spacer.test.ts`.

sample-12 "1. INTRODUCTION": **470.4pt → 445.8pt** (Word ≈ 446.2pt).

## 2. Reserve a footer taller than its bottom margin (`f3b0823`, ECMA-376 §17.10.1)

A footer is anchored at `footerTopY = pageHeight − footerDistance − footerHeight`. When the footer is taller than its bottom-margin allowance (`marginBottom − footerDistance`) its top edge rises **above** the content bottom (`pageHeight − marginBottom`) and overlaps body text. Word never lays body text over a footer, so the paginator must keep content clear of it.

sample-13's first-page masthead footer (a DOI / corresponding-author block) is ~53pt tall over a ~49pt allowance, so its top intrudes ~4pt into the content area. Fix (1) freed ~11pt on page 1, which pulled the page-2 title (`MSR - Instructions and Examples for Authors`, 18pt, no `keepNext`/`keepLines`) up onto page 1 where its descender printed over the footer.

- New `paginateWithFooterReserve`: pass 1 paginates without reservation; each page's footer is resolved and measured (`resolvePageSection` + `pickHeaderFooter` + `measureHeaderFooterHeight` at scale 1, in pt) and its overflow past the allowance, `max(0, footerHeight − (marginBottom − footerDistance))`, is fed into a second pass as a per-page content-height reservation.
- **No-op when no footer overflows** (reserve 0 — the common case), and it lives in the shared paginate path so the render worker paginates identically.
- New unit test `footer-reserve.test.ts` pins the invariant (verified red→green: it fails when the reserve is removed — body overlaps the footer by ~46pt — and passes with it).

## Verification

Browser DOM text-layer measurement (real Canvas2D render; `Range.getBoundingClientRect`, `scale = canvasWidth / pageWidthPt`):

| Sample | Check | Word ground truth | Measured | |
|---|---|---|---|---|
| sample-13 p1 | INTRODUCTION heading top | 377.6pt | **377.6pt** | ✓ unchanged |
| sample-13 p1 | page-2 title present on p1? | no | **absent (0 on-canvas hits)** | ✓ |
| sample-13 p2 | title position | top of page 2 | **70.9pt (= marginTop)** | ✓ moved cleanly |
| sample-12 p1 | "1. INTRODUCTION" top | ≈ 446.2pt | **445.8pt** | ✓ collapse intact |

Automated:

- `vitest packages/docx/src`: **295 passed** (incl. the new red→green footer-reserve test)
- `packages/node/src/docx-continuous-columns.test.ts`: **s12 = 3, s13 = 5, s5 = 7 pages**
- docx VRT: **21/21** (no regression; sample-12/13 are private and not in the VRT set)
- `tsc -p packages/docx` and root `pnpm typecheck`: clean

## Notes

- Clean-room: behaviours are reconstructed from Word's measured layout, not copied from any other implementation.
- Please do **not** squash-merge (use merge or rebase) per repo convention.

### DIFF
--- a/packages/docx/src/footer-reserve.test.ts
+++ b/packages/docx/src/footer-reserve.test.ts
@@ -1,0 +1,124 @@
+import { describe, it, expect } from 'vitest';
+import { renderDocumentToCanvas } from './renderer.js';
+import type { BodyElement, DocParagraph, DocxDocumentModel, HeaderFooter, SectionProps } from './types';
+
+// ECMA-376 §17.10.1 — a footer taller than the bottom-margin allowance
+// (marginBottom − footerDistance) rises ABOVE the content area and would overlap
+// body text. Word never lays body text over a footer: it reserves the overflow so
+// content breaks to the next page instead. The paginator measures each page's
+// footer and re-paginates with that reservation (paginateWithFooterReserve). These
+// tests pin that invariant with a synthetic doc whose first-page footer is far
+// taller than its bottom margin. Reconstructed from sample-13's masthead footer
+// behaviour (a DOI / corresponding-author block ~53pt tall over a ~49pt margin).
+
+interface Call { text: string; y: number; }
+
+function makeRecordingCanvas(): { canvas: HTMLCanvasElement; calls: Call[] } {
+  let font = '10px serif';
+  const calls: Call[] = [];
+  const ctx = {
+    get font() { return font; },
+    set font(v: string) { font = v; },
+    letterSpacing: '0px',
+    measureText: (s: string) => {
+      const p = parseFloat(/(\d+(?:\.\d+)?)px/.exec(font)?.[1] ?? '10');
+      return {
+        width: [...s].length * p * 0.5,
+        fontBoundingBoxAscent: p * 0.8, fontBoundingBoxDescent: p * 0.2,
+        actualBoundingBoxAscent: p * 0.8, actualBoundingBoxDescent: p * 0.2,
+      } as TextMetrics;
+    },
+    save() {}, restore() {}, beginPath() {}, closePath() {},
+    moveTo() {}, lineTo() {}, stroke() {}, fill() {}, fillRect() {},
+    strokeRect() {}, clip() {}, rect() {}, scale() {}, translate() {}, rotate() {},
+    setLineDash() {}, clearRect() {}, arc() {}, quadraticCurveTo() {},
+    bezierCurveTo() {}, createLinearGradient() { return { addColorStop() {} }; },
+    drawImage() {},
+    fillText(s: string, _x: number, y: number) { calls.push({ text: s, y }); },
+    strokeText(s: string, _x: number, y: number) { calls.push({ text: s, y }); },
+    fillStyle: '#000', strokeStyle: '#000', lineWidth: 1,
+    textAlign: 'left' as CanvasTextAlign, direction: 'ltr' as CanvasDirection,
+    globalAlpha: 1, lineCap: 'butt' as CanvasLineCap, lineJoin: 'miter' as CanvasLineJoin,
+  };
+  const canvas = { width: 0, height: 0, style: {} as Record<string, string>, getContext: () => ctx };
+  return { canvas: canvas as unknown as HTMLCanvasElement, calls };
+}
+
+function para(text: string): DocParagraph {
+  return {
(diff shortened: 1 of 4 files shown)


yukiyokotani:
## Independent review round → fixes (commits `5842a73`, `2bcbf6c`)

Three independent reviewers audited this branch on spec-fidelity, single-responsibility / duplication, and docx/xlsx/pptx cross-package consistency. Findings and resolutions:

### Spec-first
- **Wrong governing clause (fixed).** The footer reserve cited **§17.10.1**, which is `evenAndOddHeaders`. The actual rule is **§17.6.11 (pgMar/@bottom)**: *"the text is placed at the greater of the value of bottom / the extent of the footer text."* Re-cited throughout — and this **upgrades** the change from clean-room to spec-mandated.
- **Negative bottom margin (fixed).** §17.6.11: a negative bottom measures text from the page bottom *regardless of the footer* (it overlaps). The old `max(0, marginBottom − footerDistance)` clamp inverted this. The reserve is now the exact §17.6.11 quantity and 0 for a negative bottom.

### Single-responsibility / duplication
- **Footnote overlap (fixed).** The footnote block ignored the footer overflow, so a footnote + tall footer on one page overlapped — the same bottom band the reserve protects. The block is now raised by the footer overflow.
- **De-dup (fixed).** Footer selection and the overflow formula are extracted to `resolvePageFooter` + the unit-agnostic `footerOverflowPt`, shared by the reserve pass, the footer paint, and the footnote clearance.
- **Honesty (fixed).** The single-re-pass comment now states its assumption (exact for a page-starting first-page footer; a continuous section's tall first-page footer is the unhandled edge), and the skip threshold is a named constant.

### Collapse fix robustness
- **Paint lockstep (fixed, `2bcbf6c`).** `leadsCollapsedRun` was recomputed in paint from per-page adjacency, which can desync from the paginator at a page split. It is now stamped during pagination and read in paint — matching the existing `collapsedSpacer` pattern.

### Cross-package (docx/xlsx/pptx)
- **docx-only scope confirmed correct.** xlsx renders a fixed windowed cell grid (no footer/pagination layer — 0 footer/pagination hits in `packages/xlsx/src`); pptx renders fixed-size slides (no reflow/footer reservation). Continuous section breaks (§17.6) are docx-only. Hoisting either fix to `core` would be a false-abstraction with no second consumer — so nothing is shared.

### Verification
docx **297** unit (footnote-over-footer + negative-bottom cases added, both red→green) · VRT **21/21** · pagination s12=3 / s13=5 / s5=7 · root typecheck clean · browser DOM re-measure unchanged (s13 INTRODUCTION 377.6pt + title on page 2; s12 445.8pt).

A separate follow-up: §17.6.11's "greater of" rule is symmetric for `top`/header — a header taller than its top-margin allowance would overlap body text the same way. Not triggered by any sample; tracked separately, out of this PR's scope.

PR #600, opened 2026-06-28
fix(docx): de-listed paragraph (numId=0) drops the inherited list indent (sample-12 centered heading)

yukiyokotani:
## Summary

sample-12's centered section heading **"1. INTRODUCTION"** rendered ~18pt to the **right** of its column centre. It should be centred in its column like every other section heading.

## Root cause

`"1. Introduction"` is the document's **only** `Heading1` paragraph. Its numbering is explicitly cancelled — `<w:numPr><w:numId w:val="0"/></w:numPr>` (ECMA-376 §17.3.1.19) — and the `"1."` is **literal text**, so it is **not a list item**. Every sibling section heading (`"2. LITERATURE REVIEW"`, `"4. CONCLUSION"`, …) is a plain centred **Normal** paragraph with no indent.

`Heading1` is `basedOn` Word's built-in **List Paragraph** style (`ind left=720` = 36pt). The parser kept that inherited list indent even though numbering was removed, then the renderer correctly centred the line **within the indented box** — shifting it right by `indent/2` = 18pt. Word does not do this: removing the numbering removes the list indent, so Word centres the heading on the bare column, exactly like the Normal-styled headings.

## Fix

`packages/docx/parser/src/parser.rs`: when a paragraph's numbering is removed (effective `numId == 0`), its left/first-line indent resolves to the paragraph's **direct** `<w:ind>` only — `0` when none — instead of inheriting the list-style indent. A genuinely numbered paragraph still takes the numbering level's indent, and a direct `<w:ind>` still wins.

## Verification

| Check | Before | After |
|---|---|---|
| s12 "1. INTRODUCTION" centre | 204.7pt (+18 right) | **186.7pt** = column centre = Word PDF ("1." starts at x=137.1pt) |
| s12 "2. LITERATURE REVIEW" | 433.9pt | 433.9pt (unchanged) |
| s12 "1. INTRODUCTION" top | 445.8pt | 445.8pt (unchanged) |
| s13 INTRODUCTION (auto-numbered) | 377.6pt | 377.6pt (unchanged — numId≠0 path untouched) |

- Rust: `cargo fmt` / `clippy -D warnings` / **162 tests** (new regression test `numid_zero_drops_inherited_list_indent`, verified red→green: without the fix the de-listed indent is 36.0pt and the test fails; controls assert a numbered heading keeps 36pt and a direct `<w:ind>` survives).
- docx VRT **21/21**, docx vitest **299**, root `pnpm typecheck` clean.
- Browser DOM text-layer measurement (real Canvas2D render), not skia.

## Notes

- Clean-room: behaviour reconstructed from Word's measured output (§17.3.1.19 numbering removal). Independent of the merged PR #599 (collapse + footer reserve) and of the separate in-progress header-reserve follow-up.
- Please do **not** squash-merge (use merge or rebase) per repo convention.

### DIFF
--- a/packages/docx/parser/src/parser.rs
+++ b/packages/docx/parser/src/parser.rs
@@ -1602,8 +1602,18 @@ fn parse_paragraph_cond(
     // the mark glyph. We keep it in a separate `mark_run` used only for the
     // `default_font_*` (empty-paragraph) metrics below.
     let mut mark_run = base_run.clone();
+    // The paragraph's DIRECT indents (vs ones inherited from its style chain). A direct
+    // indent is honored even when numbering is removed (the numId=0 case below); an
+    // inherited indent is not. Captured for all three axes (start/end/first-line) so the
+    // de-list drop is symmetric for LTR and RTL.
+    let mut direct_indent_left: Option<f64> = None;
+    let mut direct_indent_first: Option<f64> = None;
+    let mut direct_indent_right: Option<f64> = None;
     if let Some(ppr) = ppr_node {
         let direct = parse_para_fmt(ppr);
+        direct_indent_left = direct.indent_left;
+        direct_indent_first = direct.indent_first;
+        direct_indent_right = direct.indent_right;
         apply_direct_para(&mut base_para, &direct);
         if let Some(rpr) = child_w(ppr, "rPr") {
             let direct_run = parse_run_fmt(rpr);
@@ -1617,7 +1627,14 @@ fn parse_paragraph_cond(
         .map(normalize_align)
         .unwrap_or("left")
         .to_string();
-    let indent_right = base_para.indent_right.unwrap_or(0.0);
+    // A de-listed paragraph (numId=0, see the indent_left/first resolution below) drops
+    // its inherited END indent too, honoring only a direct one — symmetric with the
+    // start side so an RTL list item de-lists consistently.
+    let indent_right = if base_para.num_id == Some(0) {
+        direct_indent_right.unwrap_or(0.0)
+    } else {
+        base_para.indent_right.unwrap_or(0.0)
+    };
     let space_before = base_para.space_before.unwrap_or(0.0);
     let space_after = base_para.space_after.unwrap_or(0.0);
     let line_spacing = base_para.line_spacing_val.map(|v| LineSpacing {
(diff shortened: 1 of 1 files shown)


yukiyokotani:
## Independent review round → fixes (amended into the commit)

Three independent reviewers audited this fix on spec-fidelity, correctness/edge-cases, and docx/xlsx/pptx cross-package consistency.

### Spec-fidelity (honesty) — fixed
The original comment/message cited **§17.3.1.19** as if the indent drop were literal spec. Verified against **§17.7.2** ("*the style's paragraph properties shall override the numbering level's paragraph properties*") + **§17.9.18**: `numId=0` removes only the **numbering-level** indent; List Paragraph's `ind left=720` is a **paragraph-style** property that, read literally, **survives** numbering removal (→ 36pt). Word's drop to 0 is therefore **undocumented runtime behavior**, not a literal-spec rule. The comment and commit message are relabeled as a **clean-room match of Word's measured output**, per the repo's spec-fidelity-honesty rule.

### Correctness / completeness — fixed
- **RTL symmetry** (flagged by two reviewers): the fix dropped only the start/first-line indent; an inherited **end (right)** indent persisted, so an RTL list item would de-list asymmetrically. Now drops left / first-line / **right** symmetrically (a direct `<w:ind>` still wins on every axis). Added regression cases for the symmetric right-drop and for a direct **hanging** indent surviving de-list.
- No misfire: numbered paragraphs (`numId≠0`) and never-numbered paragraphs are untouched (verified red→green and across all four private `numId=0` samples — sample-10/13 already inherit 0, so unaffected).
- Documented the deliberate over-approximation (a de-listed paragraph honors only its direct indent; all inherited indent is treated as list-derived).

### Cross-package — docx-only scope confirmed correct
pptx already drops the list indent on `<a:buNone>`, and structurally **cannot** have this bug (it never inherits `marL`/`indent` from a style chain — only the bullet is inherited, and the indent is synthesized from the effective bullet). xlsx has no list/numbering concept (`indent` is a flat `ST_CellAlignment` attribute). Nothing to hoist to `ooxml-common` (would be a false-abstraction).

### Verification
Rust `fmt` / `clippy -D warnings` / **162 tests** · docx VRT **21/21** · docx vitest **299** · root `pnpm typecheck` · browser DOM text layer (s12 "1. INTRODUCTION" **186.7pt** = column centre = Word's "1." at x=137.1pt; "2. LITERATURE REVIEW" 433.9pt and s13 auto-numbered heading unchanged).

PR #601, opened 2026-06-28
fix(docx): reserve a header taller than its top margin so body never overlaps it (§17.6.11)

yukiyokotani:
## Summary

Implements the **header-side mirror** of the footer-reserve fix ([#599](https://github.com/yukiyokotani/office-open-xml-viewer/pull/599)).

ECMA-376 **§17.6.11** (`pgMar/@top`) is the symmetric twin of the `@bottom` rule:

> If the value of top is non-negative, then the text is placed at the greater of: **the value of top** / **the extent of the header text** … the main text extent ends at the bottom of the header region.

So a header taller than its top-margin allowance (`marginTop − headerDistance`) rises into the content area and the body must start **below** it. The render path previously started the body at `marginTop` unconditionally — painting body lines over a tall header. This is the exact mirror of the overlap #599 fixed at the bottom.

## Change — mirrors the footer reserve piece-for-piece

| Footer (existing) | Header (this PR) |
|---|---|
| `footerOverflowPt` | `headerOverflowPt(headerH, marginTop, headerDistance)` |
| `resolvePageFooter` | `resolvePageHeader` |
| `computeFooterReserves` | `computeHeaderReserves` |
| `footerReservePt[]` shrinks content from bottom | `headerReservePt[]` shrinks content from top |

- `headerOverflowPt` honors the spec's **negative-top exception** (negative top margin → text measured from the page top regardless of the header, overlapping it → reserve `0`).
- `computePages` gains `headerReservePt[]` reducing `effContentH` from the top; the render path shifts the body's start `y` down by the same overflow, so pagination and paint stay in lockstep.
- `paginateWithFooterReserve` → **`paginateWithHeaderFooterReserve`** (now reserves both margins); `MIN_FOOTER_OVERFLOW_PT` → `MIN_MARGIN_OVERFLOW_PT`. Both names generalized for honesty.

## No-op for existing docs

No bundled sample triggers this (sample-13's header is short → reserve `0`). When nothing overflows, pass 1 is returned unchanged and `headerReservePt` is all-zero — an **exact no-op**, so the footer path stays byte-identical.

## Tests

Driven by `header-reserve.test.ts` (style of `footer-reserve.test.ts`): a synthetic doc whose header is far taller than its top margin asserts the first body line sits **below** the header's bottom, with a non-triviality check that the same body absent the reserve packs *into* the header band; a negative-top case reserves nothing.

- ✅ Verified **red→green** by removing the reserve, and ablated the negative-top guard (removing it fails that test).
- ✅ Full docx unit suite green (incl. `footer-reserve`, `pagination`, `page-anchor-prescan`, `docgrid-char`).
- ✅ `tsc --noEmit` clean.

docx-only, consistent with #599's cross-package adjudication (xlsx/pptx have no pagination reflow).

### DIFF
--- a/packages/docx/src/footer-reserve.test.ts
+++ b/packages/docx/src/footer-reserve.test.ts
@@ -7,7 +7,7 @@ import type { BodyElement, DocNote, DocParagraph, DocxDocumentModel, HeaderFoote
 // bottom-margin allowance (marginBottom − footerDistance) rises into the content area
 // and content (body AND footnotes) must clear it: Word never lays main text over a
 // footer. The paginator measures each page's footer and re-paginates with that
-// reservation (paginateWithFooterReserve), and the footnote block is raised by the
+// reservation (paginateWithHeaderFooterReserve), and the footnote block is raised by the
 // same overflow. A NEGATIVE bottom margin is the spec's explicit exception — text is
 // then measured from the page bottom regardless of the footer and overlaps it, so
 // nothing is reserved. These tests pin those rules with a synthetic doc whose footer
--- a/packages/docx/src/header-reserve.test.ts
+++ b/packages/docx/src/header-reserve.test.ts
@@ -15,7 +15,7 @@ import type { BodyElement, DocParagraph, DocxDocumentModel, HeaderFooter, Sectio
 // nothing is reserved. These tests pin those rules with a synthetic doc whose header is
 // far taller than its top margin (the header-side mirror of sample-13's masthead).
 
-interface Call { text: string; y: number; }
+interface Call { text: string; x: number; y: number; }
 
 function makeRecordingCanvas(): { canvas: HTMLCanvasElement; calls: Call[] } {
   let font = '10px serif';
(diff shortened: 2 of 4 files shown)


yukiyokotani:
## Independent critical review + fixes (commit d0c92d0)

Four independent agents reviewed this PR on separate dimensions — **spec fidelity**, **single-responsibility/duplication**, **cross-package consistency**, and **deep correctness/edge-cases**. Summary:

- **Spec (§17.6.11)**: faithful. Formula `max(0, headerDistance + headerH − marginTop)`, the negative-top exception, and the §17.6.11 citation were verified verbatim against the spec PDF. (One pre-existing item — body placement at `|top|` for a negative margin — is shared with the footer side and tracked as a follow-up, out of scope here.)
- **Cross-package**: reconfirmed **docx-only** by source inspection — pptx has no pagination/reflow; xlsx parses no print header/footer (`HEADER_H` there is the on-screen gutter). No missed shared-layer extraction.
- **Single-responsibility / correctness**: one **blocker** + several should-fixes, all fixed in d0c92d0:

| Finding | Fix |
|---|---|
| **BLOCKER** — multi-column: the body shift was applied to column 0 only; continuation columns (§17.6.4) reset to the bare margin top and painted into the header band | `renderBodyElements` threads the page's `headerReservePx` and re-adds it at every column-top reset |
| Page-count growth: a tall reserve grows page count, so trailing pages fell to reserve 0 and overflowed the bottom margin | `computePages` clamps a past-the-end page index to the last reserve (exact for uniform & first-page-only) |
| 8-param signature with two parallel `number[]` reserves | collapsed into one per-page `{top, bottom}` `PageReserve[]` |
| Paint inlined `pickHeaderFooter` while the reserve pass used `resolvePageHeader` (drift risk) | paint now calls `resolvePageHeader`, mirroring the footer |
| Doc-comment over-claimed; latent float gap undocumented | corrected; page-anchored-float limitation documented honestly |

**Re-verification** (adversarial): 3-column, continuous-section region-top (no double-count), and clamp edges (uniform, first-page-only `[R,0,0]`→0, empty, bottom-reserve composition) all pass. New TDD tests added (multi-column overlap; reserve-index clamp), all red→green. **301 docx unit tests pass; `tsc --noEmit` clean; CI green.**

PR #672, opened 2026-07-02
perf(core): precompiled preset formulas, bbox-sized effect canvases, region-limited bevel reads (A5+A4+A3)

yukiyokotani:
## Summary

Phase 2 items A5 / A4 / A3 from the [2026-07 improvement plan](docs/improvement-plan-2026-07.md) — the effect-pipeline hot paths, in ascending risk order, each with a byte-identity proof.

### A5 — preset-geometry formulas re-tokenized per shape per render
`evaluateFormula` split every guide expression string on every evaluation. Formulas now precompile to `{op, argTokens}` once per preset (lazy, module-level WeakMap keyed by the stable def singleton), consumed via one `evaluatorForDef` choke point for all 3 call sites; `resolve`/`applyOp` stay verbatim. **Oracle test: all 186 presets × 3 boxes resolve every guide bit-identically vs the pre-refactor evaluator. Bench: 100 shapes × 30 renders → 97,320 → 382 `split()` calls (−99.6%, ~12.9×).**

### A4 — effect aux canvases were full device size per shape
`applyInnerShadow`/`applySoftEdge`(×3)/`applyReflection` allocated deviceW×deviceH canvases per shape — the passed `effBBox` was ignored. They now crop to `bbox ⊕ margin`, painting through a small transform-composing proxy and blitting back at the crop origin. **The real-skia pixel test caught a genuine subtlety: CSS `blur(r)` treats r as the std-dev, so the kernel reaches ~3σ — margins are 3·blur** (softEdge keeps `radius`). 6 pixel-equality cases (interior/edge-clamped/±blur) all byte-exact vs a verbatim full-canvas oracle. **Pixels touched: −~99% (79–87×)** for a 400×240 shape on a 4K canvas. Pool deliberately not adopted — bbox-sized dimensions rarely repeat (documented).

### A3 — bevel/extrusion read the whole canvas
`applyBevelShading`/`applyExtrusion` now accept an optional region limiting `getImageData`/loop/`putImageData`; pptx passes the body box grown by band width/offset. Byte-identical under region-equivalence tests (incl. edge-clamped). **Honest caveat, documented in-code**: pptx already gives each 3D shape its own shape-sized offscreen, so today's win there is just the transparent pad — the seam pays off for any larger-canvas caller, and GPU compositing was deliberately not attempted (VRT-divergence risk).

## Verification
- `pnpm test` 1555 (+9) / typecheck / lint — green
- **VRT all formats: pptx 75 / docx 21 / xlsx 67 — all pass**, incl. the dedicated bevel-ring spec 6/6 (sample-11 slide-6 sp3d ring — exactly the A3 case — at devScale 1/2/2) and worker-equivalence 0.000%
- Browser smoke delegated to CI (worktree port rule)

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -159,6 +159,7 @@ export {
   applyBevelShading,
   applyExtrusion,
   type ExtrusionInput,
+  type BevelRegion,
   computeBevelNormals,
   bevelHeightProfile,
   distanceToEdge,
--- a/packages/core/src/shape/bevel-shading.test.ts
+++ b/packages/core/src/shape/bevel-shading.test.ts
@@ -804,6 +804,133 @@ describe('applyBevelShading (end-to-end on a filled square)', () => {
   });
 });
 
+/**
+ * A region-honouring ImageData-backed ctx: `getImageData(sx,sy,sw,sh)` returns a
+ * copy of exactly that window, and `putImageData(img,dx,dy)` writes it back at the
+ * offset — so it exercises the A3 sub-region path (the simple `fakeCtx` above
+ * ignores the coordinates, which is fine for the full-canvas callers). Tracks the
+ * getImageData window so the test can assert the region actually shrank.
+ */
+function regionCtx(
+  w: number,
+  h: number,
+  fill: (x: number, y: number) => [number, number, number, number],
+): BevelCtx & { data: Uint8ClampedArray; lastGet: { x: number; y: number; w: number; h: number } | null } {
+  const data = new Uint8ClampedArray(w * h * 4);
+  for (let y = 0; y < h; y++)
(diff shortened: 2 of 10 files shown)


yukiyokotani:
CI failure root-caused: the reflection blit **resamples** (mirror transform carries a fractional translation), so skia's bilinear sampling near the cropped source edge is platform-dependent — macOS matched by luck, Linux diverged by ≤7/255. innerShadow/softEdge blit with integer-offset identity transforms (pure pixel copy, byte-exact by construction) and stay cropped. `applyReflection` is reverted to the verbatim full-canvas path with an in-code WHY note, and the two reflection pixel tests are re-labeled as tripwires (identical-code comparison — they fail if anyone re-crops reflection without solving the resample problem). a051dc9

PR #684, opened 2026-07-02
refactor(docx): paginator's layout lines are reused by paint — compute-once stage 1 (B2)

yukiyokotani:
## Summary

Phase 4-1 B2 **Stage 1** from the [2026-07 improvement plan](docs/improvement-plan-2026-07.md): the paginator computed `layoutLines` for every paragraph, kept only `lineSlice`, and threw the lines away — paint recomputed everything. First step of retiring the measure/paint double bookkeeping.

### Scale-linearity characterization (Stage 2's design input)
Under a linear mock font `layoutLines` is **perfectly scale-linear in every field** (px fields ×scale; structure invariant; grep-verified no px quantization in the function body). **The only scale-dependent thing is the line partition itself**, driven by real-font `measureText` hinting non-linearity (documented by a sub-linear-mock test). Everything else rescales cleanly — that's the exact boundary Stage 2 must handle.

### Compute-once with a self-verifying gate
The paginator now stamps its scale-1 lines next to `lineSlice`; `renderParagraph` reuses them **only when** `scale === 1` **and** every layout input it would pass matches what the paginator recorded (paraW / firstIndent / tabOrigin / gridDelta / no-float) — pixel-exact by construction, no rescaling attempted. The default `renderPage` path renders at scale 1, so the gate fires on the common path. Numbered lists deliberately fail the gate: the known measure-vs-paint `firstLineIndent` drift (`para.indentFirst` vs `numBodyOffset`) is real, confirmed, and left untouched — unifying it is Stage 2's explicit behaviour change.

### Proof & payoff
- Temporary real-Chromium spec rendered every docx sample page reuse-ON vs forced-OFF through the full cross-context path: **diffPixels = 0 on all 19 pages** (spec reverted after use).
- VRT all formats: no baseline change (docx 21 / pptx 75 / xlsx 67).
- Re-render bench (53-page doc): **~2.2× faster** (~10.3 → ~4.7 ms/full-doc; wrap-loop measureText skipped on reused paragraphs) — the B4 payoff, delivered structurally.

## Verification
- pnpm test 1603 / typecheck / lint — green
- VRT all formats — no baseline change
- Browser smoke delegated to CI

### DIFF
--- a/packages/docx/src/layout-lines-reuse-identity.test.ts
+++ b/packages/docx/src/layout-lines-reuse-identity.test.ts
@@ -117,7 +117,7 @@ function para(text: string, over: Partial<DocParagraph> = {}): DocParagraph {
   } as unknown as DocParagraph;
 }
 
-function doc(body: BodyElement[], pageHeight = 60): DocxDocumentModel {
+function doc(body: BodyElement[], pageHeight = 60, settings?: Record<string, unknown>): DocxDocumentModel {
   const section: SectionProps = {
     pageWidth: 200, pageHeight,
     marginTop: 10, marginRight: 10, marginBottom: 10, marginLeft: 10,
@@ -130,6 +130,7 @@ function doc(body: BodyElement[], pageHeight = 60): DocxDocumentModel {
     footers: { default: null, first: null, even: null },
     fontFamilyClasses: { 'Times New Roman': 'roman' },
     footnotes: [],
+    ...(settings ? { settings } : {}),
   } as unknown as DocxDocumentModel;
 }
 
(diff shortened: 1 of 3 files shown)


yukiyokotani:
Review response (78c211a): (1) kinsoku added to the gate — `===` was proven non-viable (resolveKinsokuRules builds fresh Sets per call and the prebuiltPages path resolves twice), so a value-equivalence check compares enabled + both forbidden sets. (2) font-loading contract documented on paginateDocument. (3) The segs-stability premise was **false** — page/NUMPAGES fields and noteRef labels are state-dependent, so the original commit could have painted a stale page count under reuse; a `paragraphSegsStateSensitive` predicate now skips stamping for such paragraphs (recompute path preserved), with load-bearing tests that go red if either fix is removed. docx suite 462 pass, VRT diff figures byte-identical.

Issue #754, opened 2026-07-04
docx: sample-25 table fidelity — header shading/alignment lost, 'Production' clipped


yukiyokotani:
## Summary
The 2-column table in `private/sample-25.docx` (user-authored Word file, below the pie chart) renders with multiple fidelity breaks vs Word's PDF (`private/sample-25.pdf`):

1. **Header clipped mid-glyph**: "Production" renders as "**oduction**".
2. **Header row shading lost**: Word shows a **blue header row** (built-in table style firstRow shading, white text); we render white background.
3. **Header alignment**: both "Country" and "Production" headers should be **centered**; ours are not.
4. **Data alignment**: "Country" column data should be **left-aligned**, "Production" column data **centered**; ours differ.

(User-verified against the PDF, 2026-07-05.)

## Suspects
Built-in table-style conditional formatting resolution (`tblStyle` + `firstRow`/banded `tblStylePr` — shading, jc, and cell margins from the style layer), plus column width / text-start offset for the clipping. Possibly related to the parked V-1 cnf-precedence question (#514–530 batch).

## Note
Fixture is local-only (private, not redistributable); reproduce with the user's sample-25.docx + PDF.

PR #766, opened 2026-07-04
fix(core,docx,pptx,xlsx): preserve trailing-space advance on trimming measureText engines

yukiyokotani:
## Summary
A run like `<w:t>Other countries</w:t>` rendered as "Othecountries" in sample-25's table — the inter-word space advance was dropped. Root cause is **not** bold or table styling: `measureText` for a string **ending in whitespace** is engine-dependent — Chrome/Blink and skia-canvas include the trailing-space advance; **Firefox and WebKit exclude it**. Word-split segments keep their trailing space (`["Other ", "countries"]`), so on a trimming engine the measured advance omits the space and the next word collides. All three packages had the hole.

## Fix
New shared core helper `measureAdvanceWithTrailingSpace(ctx, text)` (`core/src/text/measure-space.ts`): detects once per (context, font) whether the engine trims (interior `"m m"−"mm"` vs trailing `"m "−"m"`) and only then adds back `trailingSpaceCount × spaceAdvance`. **Pure pass-through on non-trimming engines** → zero behavior change on Chrome/skia.

- docx (`d9a439b`): `layoutLines` + `rescaleLayoutLines` measure sites (both needed or draw disagrees with pagination) + tests.
- pptx (`37a09d6`): whitespace tokens measured as lone `" "` (fully trimmed to 0 on trimming engines).
- xlsx (`66a42f0`): pen-chained rich-run/wrap segments ending at a space. Whole-cell single-fillText paths unaffected (interior spaces never at a measured edge) — left as-is.

## Verification
- sample-25 with a patched trimming `measureText`: before `"Other "` w=39.1 → collision (exact reported symptom); after w=43.17 → space restored, identical to native-Chrome positions.
- Native Chrome & skia: byte-identical before/after.
- vitest core+docx+pptx+xlsx **2084 pass** (+9 new: engine-mock unit tests + end-to-end collision guard); node 34; tsc/lint clean.
- Full VRT: docx 21 / pptx 12 / xlsx 8 pass, 0 failures, **no reference updates** (pass-through on the VRT engine).

Found via a background-task chip during the #754 table-style work (pre-existing on main).

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -292,6 +292,11 @@ export {
   type SegStretch,
 } from './text/line-distribute';
 export { justifiedPiecePositions, type JustifiedPiece } from './text/justify-positions';
+// Trailing-whitespace-robust text advance: `measureText` excludes a run's
+// trailing/lone whitespace on Firefox & WebKit/Safari but includes it on
+// Chrome/Blink & skia-canvas. Shared so docx/pptx/xlsx keep inter-word spaces
+// from collapsing on trimming engines (measure==draw on every engine).
+export { measureAdvanceWithTrailingSpace } from './text/measure-space';
 // Format-agnostic index navigation for hidden-item "skip" mode (pptx hidden
 // slides, xlsx hidden sheets): pure math over an isHidden(i) callback.
 export { nextVisibleIndex, resolveVisibleIndex, countVisible } from './nav/visible-index';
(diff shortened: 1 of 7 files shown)


yukiyokotani:
Closing after independent review: empirical probing of real WebKit 26.4 / Firefox 148 / Chromium / skia-canvas shows **no current engine trims trailing spaces from measureText** — the helper is a no-op everywhere (pixel-identical renders proven), and the observed sample-25 'Othecountries' collision reproduces byte-identically with and without this change. The real root cause is a **canvas state leak in core's renderChart()** (textAlign=center / textBaseline=middle leak from chart label painting into all subsequent same-page text) — being fixed separately with a save/try/finally-restore in core plus a regression test. The branch is kept in case a trimming engine ever ships.

PR #808, opened 2026-07-05
fix(charts): place solid-pie ctr data labels near the rim (§21.2.2.48)

yukiyokotani:
## Problem

On sample-14 slide-7, PowerPoint places the solid pie's percent labels
(`54% / 27% / 14% / 5%`, chart6, `<c:dLblPos val=\"ctr\"/>`) **near the outer
rim**, but the renderer put them at **half the radius** (mid-disc). The
neighbouring doughnut (chart7) was already correct.

## Root cause

`drawPieRichLabels` positioned every non-`outEnd` label at the ring midpoint
`(innerR + outerR) / 2`. That is right for a doughnut but, for a **solid pie**
(`innerR = 0`), collapses to `0.5·outerR`.

## Ground truth (measured from sample-14.pdf)

Center + outer radius come from a least-squares fit to the pie's outer rim
(residual std 0.43pt). Label centroids:

| slice | measured r / R |
|-------|----------------|
| 54%   | 0.878 |
| 27%   | 0.888 |
| 14%   | 0.887 |
| 5%    | 0.912 |

A flat near-rim constant, independent of slice angle — so it is a fixed
fraction, **not** a sector centroid. The doughnut (55% hole) measures
0.772–0.778·outerR ≈ `(0.55 + 1)/2 = 0.775`, i.e. the exact ring midpoint.

§21.2.2.48 enumerates `ST_DLblPos` (bestFit / ctr / inEnd / outEnd …) but
defines **no geometry** for these — the radial placement is PowerPoint
application behaviour. Reverse-engineering was explicitly requested for this
fidelity match; the measurement method and cross-slice consistency are
documented at the new constant and at the `labelR` computation.

## Fix

- Solid pie (`innerR ≈ 0`) → `ctr` / `inEnd` / `bestFit` labels at
  `0.88·outerR` (`PIE_CTR_LABEL_RADIUS_FRAC`).
- Doughnut (`innerR > 0`) → unchanged ring midpoint `(innerR + outerR)/2`.
- `outEnd` (outside the rim) → unchanged.
- Legacy inline percent path (plain pie without rich `<c:dLbls>`) → untouched,
  so the renderer-signature snapshot stays byte-stable. sample-25 pie3D uses
  the callout path and is unaffected.

## Verification

- **TDD**: added `renderer-scatter-line-and-pie-labels.test.ts` cases asserting
  the pie `ctr` label radius (~0.88R, Red→Green), the doughnut ring-midpoint
  (0.775R), and `outEnd` outside the rim.
- **End-to-end (real pptx pipeline)**: sample-14 slide-7 pie label radius ratio
  moved **0.50 → 0.88** (matching the PDF); the doughnut right-half render hash
  is **byte-identical** before/after (`9be1c143…`).
- `renderer-signature` snapshot unchanged; core chart vitest (261) green; core
  package vitest (1152) green; `tsc` (core) clean; `ast-grep scan` no new
  violations; pptx/xlsx demo VRT green (private-fixture 404s are expected in
  this worktree).

### DIFF
--- a/packages/core/src/chart/renderer-scatter-line-and-pie-labels.test.ts
+++ b/packages/core/src/chart/renderer-scatter-line-and-pie-labels.test.ts
@@ -63,19 +63,23 @@ function series(over: Partial<ChartSeries>): ChartSeries {
   return { name: '', color: null, values: [], ...over };
 }
 
-interface TextCall { text: string; fillStyle: string }
+interface TextCall { text: string; fillStyle: string; x: number; y: number }
 
 /** Recording context that captures: (a) `stroke()` calls that follow a
  *  `moveTo`/`lineTo` path with ≥2 vertices AND whose strokeStyle is `matchColor`
  *  (the series line color) — this isolates the scatter connecting line from
  *  gridlines/axis rules, which stroke in grey; (b) `fillText` calls with the
  *  fillStyle in effect. Enough to assert the scatter line pass and pie labels. */
+interface ArcCall { cx: number; cy: number; r: number }
+
 function recordingCtx(matchColor = '#4f81bd'): {
   ctx: CanvasRenderingContext2D;
   counts: { polylineStrokes: number };
   texts: TextCall[];
+  arcs: ArcCall[];
 } {
   const texts: TextCall[] = [];
+  const arcs: ArcCall[] = [];
   const counts = { polylineStrokes: 0 };
   let pathVerts = 0;
   const state: Record<string, unknown> = {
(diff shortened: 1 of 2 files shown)


yukiyokotani:
Review notes for the record: (1) the PR body's chart-test count should read 258, not 261 (3 new cases confirmed present and passing). (2) Pre-existing latent inconsistency left deliberately untouched: the legacy inline pie-label path (renderer.ts ~L2724) still places labels at 0.6R while the rich-dLbls path now uses 0.88R — no current sample exercises the legacy path and no ground truth exists for it; unify if such a sample ever surfaces.

Issue #771, opened 2026-07-04
docx: vertical-text stage-2 — spec-driven glyph positioning + remaining tbRl features


yukiyokotani:
Follow-up to the vertical-writing stage-1 work (PR #770, §17.6.20 tbRl).

Stage-1 ships the page rotation + upright-CJK glyph path, anchor/inline/float image uprighting, and the text-selection overlay transform, all verified against `private/sample-26.docx` / `sample-26.pdf` (image centroid matches Word/PDF within ~1px). It carries a few **explicitly-flagged heuristics** and deferred features to replace with spec-driven implementations:

### Heuristic constants to remove (flagged in code as stage-1 approximations)
- `packages/docx/src/vertical-text.ts` `drawVerticalRun`: the `+0.12em` baseline nudge for centering upright CJK glyphs is font-dependent and NOT a spec constant. Replace with placement from the font's vertical/ideographic metrics (e.g. `textBaseline='ideographic'` once broadly supported, or measured em-box).
- `packages/docx/src/vertical-text.ts` `verticalGlyphOffset`: the `{dx:0.4, dy:-0.4}` em nudge for the small comma/period (、。，．) is not from JIS X 4051. Replace with **vertical-form glyph substitution** (U+FE10–U+FE19 CJK Compatibility Forms) so the font supplies the pre-positioned vertical punctuation; then delete the offset.

### Deferred tbRl features (currently rendered as horizontal or approximated)
- `btLr` text direction (Strict `lr`): vertical but lines flow left→right / glyphs stack bottom→top — needs the opposite page rotation/flow. Currently parsed but `isVerticalSection` returns false (renders horizontal).
- 縦中横 (tate-chū-yoko): short Latin/digit runs set upright and combined horizontally within a vertical line.
- Paragraph-relative `positionV` anchors in tbRl: `resolveAnchorBox` maps page/margin-relative anchors physically; a `relativeFrom="paragraph"` vertical anchor's paraBaseY is a logical flow coord and is not physical-mapped yet.
- Vertical tables: the cell Y-band clip uses `ctx.canvas.width` (physical) → over-wide in logical space (harmless, but untested); header/footer in tbRl (`verticalLayoutSection` maps header/footer distances best-effort, unexercised).
- Vertical line-break/kinsoku and docGrid pitch semantics specific to vertical flow (stage-1 reuses the horizontal engine on swapped geometry).

All heuristic sites are commented in-code referencing this issue.

[...]

PR #834, opened 2026-07-05
fix(docx): render 縦中横 (eastAsianLayout vert) as a one-cell horizontal group (§17.3.2.10)

yukiyokotani:
## Summary

Implements **ECMA-376 §17.3.2.10 `<w:eastAsianLayout w:vert>`** (縦中横 / *tate-chū-yoko*, horizontal-in-vertical) for DOCX vertical (tbRl) pages, and **fixes a regression from #813** where `sample-26`'s date "９月２９日" rendered with the "９" digit and "日" overlapping.

### Root cause of the regression

The "２９" run carries `<w:w w:val="67"/>` **and** `<w:eastAsianLayout w:id="…" w:vert="1" w:vertCompress="1"/>`. PR #813 folded `w:w=67` into the *measured* advance, shrinking the "２９" cell to 67% along the vertical column — but the two full-width digits were still painted full-width, upright, stacked vertically, so they overran the compressed cell and collided with the following "日". The real fix is that `w:eastAsianLayout w:vert` (縦中横) was never parsed: Word lays "２９" out as **one cell of the vertical line with the digits side by side horizontally** (verified against `sample-26.pdf`: exactly one 12 pt cell).

### What changed

**Parser (§17.3.2.10)** — parse `vert` / `vertCompress` (drive the renderer) and `combine` / `combineBrackets` (carried for a future two-lines-in-one implementation; no fixture yet) into the run model, resolved through the style chain like the other run properties. Adds an `on_off_attr` helper for ST_OnOff-valued *attributes* (§22.9.2.7), distinct from `bool_prop` (child-element `w:val`).

**Renderer / layout (measure == paint)**
- A 縦中横 seg advances **exactly one em (one cell)** along the column (`segAdvanceWidth` short-circuits), independent of char count and `w:w`.
- `buildSegments` folds in the vertical gate — §17.3.2.10 is meaningful **only** in vertical text — so the flag is inert on horizontal pages (they stay byte-identical).
- `drawTateChuYokoRun` draws the whole run as one upright, centred `fillText` across the column; `w:w` (§17.3.2.43) compresses the digits' **cross-column width** via `scale(charScale, 1)`; `w:vertCompress` fits their **height** to one em (no-op for single-line content).

### §17.3.2.10 composition rule (vert / vertCompress / w:w)

| axis | attribute | effect |
|---|---|---|
| along-column (cell height) | `vert` | run occupies **one cell** (one em), regardless of char count |
| cross-column (glyph width) | `w:w` | digits laid side by side, scaled — allowed to overflow the column width |
| along-column (height fit) | `vertCompress` | squeeze the rotated run's height to one line (no-op when it already fits) |

### "9月29日" spacing (along-column advance, pt)

| cell | `sample-26.pdf` (ground truth) | before (main) | after (this PR) |
|---|---|---|---|
| ９ | 12.0 | 12.3 | 12.1 |
| 月 | 12.0 | 12.3 | 12.1 |
| **２９** | **12.0 (one cell)** | **16.7 (overflows → overlaps 日)** | **12.2 (one cell)** |

The "２９" cell now advances one em, the two digits sit side by side inside it, and "日" abuts the next cell — matching Word.

### Verification

- **Real-browser render of `sample-26.docx`** (Chrome, actual CJK shaping): "２９" now a single 縦中横 cell, "９"/"日" no longer overlap — visually matches `sample-26.pdf`. Before/after screenshots captured locally.
- New unit tests: `drawTateChuYokoRun` (vert-only / vert+w:w / vert+vertCompress / tall-content compression / 3-digit combined), `segAdvanceWidth` one-em pin.
- New end-to-end render test pins the "9月29日" no-overlap invariant and the horizontal-page byte-identical inertness of the flag.
- Gates: `cargo test/fmt/clippy -D warnings` (277 pass), docx `vitest` (803 pass) + `tsc`, `ast-grep` (0 new findings), docx demo VRT byte-identical.

### Scope / follow-ups
- `combine` / `combineBrackets` (two-lines-in-one) are parsed but not yet drawn (no fixture) — flagged inline.
- Full-width digits keep the run as one segment; a 縦中横 run mixing EA and Latin content would split per-slot — not present in any fixture.
- The `sample-26` vertical VRT reference is private/gitignored and **not** updated here (would have captured the buggy render).

Refs #771.

### DIFF
--- a/packages/docx/parser/src/parser.rs
+++ b/packages/docx/parser/src/parser.rs
@@ -2943,6 +2943,15 @@ fn parse_run_inner(
     let char_scale = fmt.char_scale;
     let position = fmt.position;
     let kerning = fmt.kerning;
+    // East Asian typography (ECMA-376 §17.3.2.10 eastAsianLayout), resolved
+    // through the style chain into `fmt`. `east_asian_vert` (縦中横) drives the
+    // vertical (tbRl) renderer's horizontal-in-vertical cell; `vert_compress`
+    // fits it to the line height. `combine`/`combine_brackets` are carried for a
+    // future two-lines-in-one implementation (parsed, not yet drawn).
+    let east_asian_vert = fmt.east_asian_vert;
+    let east_asian_vert_compress = fmt.east_asian_vert_compress;
+    let east_asian_combine = fmt.east_asian_combine;
+    let east_asian_combine_brackets = fmt.east_asian_combine_brackets.clone();
 
     // Set by the "noBreakHyphen" arm below when it just pushed/extended a text
     // run for a `<w:noBreakHyphen/>` — tells the VERY NEXT loop iteration's
@@ -3000,6 +3009,10 @@ fn parse_run_inner(
                         char_scale,
                         position,
                         kerning,
+                        east_asian_vert,
+                        east_asian_vert_compress,
+                        east_asian_combine,
+                        east_asian_combine_brackets: east_asian_combine_brackets.clone(),
                         note_ref: None,
                     };
                     match runs.last_mut() {
@@ -3073,6 +3086,10 @@ fn parse_run_inner(
                         char_scale,
                         position,
                         kerning,
+                        east_asian_vert,
+                        east_asian_vert_compress,
+                        east_asian_combine,
+                        east_asian_combine_brackets: east_asian_combine_brackets.clone(),
                         note_ref: None,
                     })));
                 }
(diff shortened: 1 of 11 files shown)


yukiyokotani:
Review note for the record: Word's actual ink advance around the 縦中横 cell measures ~13.7-13.9pt (pdftotext bbox) vs the renderer's uniform ~12.1pt em pitch — Word appears to widen the cell slightly to absorb the 15.65pt ink. The one-cell model is spec-correct and the overlap is gone; the ~1.7pt/boundary tighter advance is a minor fidelity residual tracked within the #771 stage-2 family.

PR #844, opened 2026-07-05
fix(docx): resolve column-relative anchors against the text column (§20.4.3.4)

yukiyokotani:
## Problem

In sample-28 (Arabic RTL letter) the Vision-2030 / Ministry-of-Justice **header logo** sits too far right, its right emblem crowding the page edge instead of matching Word's PDF (ground truth `sample-28.pdf`).

## Root cause (ECMA-376 §20.4.3.4)

The logo is a `wrapNone` floating anchor:

```xml
<wp:positionH relativeFrom="column"><wp:posOffset>5091023</wp:posOffset></wp:positionH>  <!-- 400.9 pt -->
<wp:extent cx="2068803" cy="470424"/>                                                   <!-- 162.9 pt wide -->
```

authored inside the **second `<w:tc>`** of a `bidiVisual` header table.

`xContainer` degraded `relativeFrom="column"` to the **page margin band**, so the 400.9 pt offset was measured from the page's left text margin (21.3 pt) rather than from the **containing cell's text column**. ST_RelFromH §20.4.3.4 defines `column` as *"relative to the extents of the column which contains its anchor"* — the current text column, not the page margins.

## Bisection — NOT a #827 (`w:tblInd`) regression

The task flagged that #827 (`b3bd2f4`, negative `tblInd`) reshaped this header table's origin ([21..522]→[14..588]). I built and rendered both revisions:

| revision | logo green-ink x (px, W=595) |
|---|---|
| pre-#827 `82bd1e8` | [437 .. 578] |
| current main `e1ec761` | [437 .. 578] |

**Identical** — the logo never moved with #827. The mis-placement predates `tblInd`; the image resolves page-margin-relative (not table-relative) on both. #827's header rule reaching the page edge (`[14..588]`) is preserved by this change.

## Fix

`column` (and `character`, which has no run-relative offset data and degrades to the same containing column — the closest available base) now resolve against `state.contentX` / `state.contentW`, which the renderer already tracks as the current text column:

- **body, single column** → the section margin band (**unchanged**: `contentX == marginLeft` there, so body-level `column` anchors — e.g. sample-13's journal masthead — are byte-identical);
- **multi-column section** → the specific column band;
- **table cell** → the cell's inner text box (`renderCell` sets `contentX = cellX + cellMarginLeft`).

## Measured — matches Word's PDF (±1 pt)

Full logo ink (right-half, px→pt at width 1000):

| | left | right | right-edge clip |
|---|---|---|---|
| Word PDF (ground truth) | 356.9 | 587.1 | — |
| **after fix** | 357.3 | 588.3 | none |

Green-ink logo extent: `[437..578]` → `[434..577]` vs PDF `[433.8..577.0]` (≤0.3 pt).

## Verification

- New `anchor-column.test.ts` (TDD): pins `column`/`character` → contentX/contentW, `margin`/`page` unchanged, offset-from-column-left, and body == margin.
- All **809** docx unit tests pass; `tsc --build` clean; ast-grep clean.
- **docx demo VRT** (`demo/sample-1`, the only committed reference) byte-identical — the private `sample-N` failures are `ERR_CONNECTION_REFUSED` on absent gitignored fixtures, not pixel regressions.
- **Cross-package**: `column` is a WordprocessingML-only anchor base (`wp:` namespace); pptx (slide-absolute `a:off`) and xlsx (cell anchors) have no equivalent path, so no shared穴 to fix.


yukiyokotani:
## Review F1 addressed — `ecf96b0`

**Root cause confirmed as reported**: `buildMeasureState` seeded `contentX: 0` while the paint pass seeds `contentX = marginLeft × scale`. Inert until the §20.4.3.4 change made `xContainer` consume it — the pagination measure pass then resolved body-level `column` anchors a full `marginLeft` left of the paint pass (`c.start = 0` vs `72`), shifting float wrap bands only during measurement (sample-9 p3–10 drift).

**Fix**: seed `contentX: section.marginLeft` (scale = 1 in the measure pass), mirroring the paint-side `baseState`. `contentW` already matched.

### Measured (page sha256 vs main `e1ec761`, width 595, dpr 1)

| sample | pages | result |
|---|---|---|
| sample-9 | 11 | **all byte-identical to main** (regression gone) |
| sample-10 | 2 | byte-identical — the previously observed drift was collateral of this same measure-seed bug and disappears with it |
| sample-12 / 13 / 17 / 18 / 21 | 3 / 5 / 3 / 3 / 2 | all byte-identical |
| sample-28 | 20 | all pages change as intended (header logo on every page) |

### sample-28 effect maintained

- p.1 logo green-ink `x=[434.1..577.0]` vs Word PDF `[433.8..577.0]` (≤0.3 pt); full right-half ink `[357.3..588.3]` vs `[356.9..587.1]` (~1 pt), no right-edge clip.
- p.2+ belong to the `marginLeft=72pt` section, where the old margin-based resolution put the logo at `72+400.9+162.9 = 635.8pt > 595pt` page width — a real clip (visible on p.5). Cell-relative resolution is section-stable, so the clip stays resolved.

### F4: paginator-level pin added

`paginate-column-anchor.test.ts` drives **`computePages`** with a body-level column-anchored square float and asserts it paginates identically to a margin-relative control (at body level the column band == margin band). Under the buggy seed the float shifts `marginLeft` left, invades the text band, and splits the paragraph (`lineSlice {0,7}` appeared — verified red before the fix). This covers the measure-pass interaction the hand-constructed MinState units cannot reach.

### Gates

810 docx vitest ✓ · `tsc --build` ✓ · ast-grep ✓ · docx demo VRT (demo/sample-1 ×6 + worker-equivalence) ✓

PR #864, opened 2026-07-06
fix(xlsx): resolve implicit cell references in sparkline range extraction

yukiyokotani:
## Summary

`extract_range_values` (the sparkline data-range extractor, near `MAX_SPARKLINE_CELLS`) scanned `<c>` elements with a flat `doc.descendants()` walk and **silently skipped any cell without an explicit `@r`** (`let Some(r_attr) = c.attribute("r") else { continue; }`). ECMA-376 marks `@r` `use="optional"` on both `CT_Cell` (§18.3.1.4) and `CT_Row` (§18.3.1.73), so a sparkline whose source worksheet uses the minimal r-less form resolved to no data and rendered blank. Because the scan was flat, it had no row grouping and no running column, so it could not infer an omitted position.

**Relation to #851**: PR #851 taught the *main* cell path (`parse_worksheet` row loop / `parse_row_cells`) to resolve omitted references by ordinal document order, but `extract_range_values` was left on the old flat scan. This PR is the mirror fix for the sparkline data path.

## Fix

`extract_range_values` now walks `<row>` elements first:

- **row** without `@r` → previous row + 1 (first row = 1); an explicit `@r` re-anchors the counter (§18.3.1.73).
- **cell** without `@r` → previous cell's column + 1 within that row (first cell = column A); an explicit `@r` re-anchors the running column and, authoritatively, the row (§18.3.1.4).

### Shared helper

Both #851 sites and this new path reduce to the same primitive: *resolve an optional 1-based ordinal, defaulting to previous + 1, and re-anchor on an explicit value.* Extracted as a small pure fn `resolve_implicit_ordinal(explicit, &mut prev)` and routed through by all three consumers (`parse_worksheet` row numbers, `parse_row_cells` cell columns, `extract_range_values` sparkline data cells) so their implicit-reference handling cannot drift. The abstraction is minimal (type + running-counter logic only); the surrounding loop shapes remain distinct, so no false abstraction is forced.

## TDD

5 new unit tests in `sparkline_range_cap_tests`, all failing before / passing after:

- `all_implicit_refs_resolve_row_major` — row & cell both r-less → A1,A2,A3.
- `implicit_cells_fill_columns_left_to_right` — consecutive r-less cells → A1,B1,C1.
- `explicit_ref_reanchors_running_counters` — mixed explicit/implicit `<c>` re-anchoring + implicit row after explicit `r="1"`.
- `explicit_row_ref_reanchors_row_counter` — implicit row after explicit `r="5"` → row 6.
- `implicit_cells_respect_type_filter` — a `t="s"` cell in an implicit run maps to None but still advances the column counter.

Red output (pre-fix), e.g. `explicit_ref_reanchors_running_counters` produced `[None, None, Some(7.0), None, None, None, None, None]` — only the single explicit `C1` cell survived; every implicit cell was dropped.

## Gates

- `cargo test` — **148 passed** (140 pre-existing + 8 in `sparkline_range_cap_tests`), 0 failed. The 3 pre-existing cap tests still pass; #851's own suite unaffected by routing its two sites through the shared helper.
- `cargo fmt --all --check` — clean.
- `cargo clippy --all-targets -- -D warnings` — clean.
- No `Cargo.lock` diff.
- WASM rebuilt via `pnpm --filter @silurus/ooxml-xlsx wasm` (reinit injected).
- xlsx vitest — **503 passed** (54 files).
- xlsx demo VRT (`demo/sample-1`, 5 sheets) — 100% match, worker-vs-main 0.000%. Reference images untouched.

## Note on severity

Real-world producers always emit explicit `@r`, so this is **spec-completeness** for the optional minimal r-less form (emitted by some enterprise exporters), not a field-reported failure.

### DIFF
--- a/packages/xlsx/parser/src/lib.rs
+++ b/packages/xlsx/parser/src/lib.rs
@@ -1244,11 +1244,12 @@ fn parse_worksheet(
                 // (implicit sequential numbering — the de-facto consumer
                 // convention; the spec only grants the optionality). An explicit
                 // value also re-anchors the counter for following implicit rows.
-                let row_idx: u32 = match node.attribute("r").and_then(|s| s.parse::<u32>().ok()) {
-                    Some(r) => r,
-                    None => prev_row_idx + 1,
-                };
-                prev_row_idx = row_idx;
+                // Routed through the shared primitive so this and the sparkline
+                // data path (`extract_range_values`) cannot drift.
+                let row_idx = resolve_implicit_ordinal(
+                    node.attribute("r").and_then(|s| s.parse::<u32>().ok()),
+                    &mut prev_row_idx,
+                );
                 let hidden = attr_bool(&node, "hidden").unwrap_or(false);
                 // ECMA-376 §18.3.1.73 `<row>@ht` is the row height in points.
                 // Gating the value on `@customHeight="1"` (0.37.0) was too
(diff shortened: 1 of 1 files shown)


yukiyokotani:
Review correction (orchestrator's task description error, surfaced by the adversarial review): the byte-identical invariance check applies to the actual sparkline fixtures **sample-7.xlsx and sample-25.xlsx** (sample-30 has charts, not sparklines). Verified MD5-identical parse dumps on both plus demo/sample-1, old vs new WASM. The implicit-ref branch ships with synthetic-only coverage (6 unit tests + 4 review probes) since all real-world producers emit explicit @r.

PR #866, opened 2026-07-06
fix(pptx): apply local envelope slope shear to warped WordArt glyphs

yukiyokotani:
## Problem

`private/sample-16.pptx` slide 1 "Wave One" (`prstTxWarp prst="textWave1"`) did not match Word/PowerPoint. Glyph **positions** were right, but glyphs stood **upright** — Word shears them so the letters lean italic-forward and lie along the wave (`packages/pptx/public/private/sample-16.pdf`, the user's Office-exported ground truth).

## Root cause (ECMA-376 §20.1.9.19, `prstTxWarp`)

`renderWarpedText` (`packages/pptx/src/renderer.ts`) placed each glyph via `warpGlyphTransform` (`packages/core/src/shape/text-warp.ts`) as

```
translate(x,y) · rotate(angle) · scale(hScale, vScale)   // angle = mean edge tangent
```

a **rigid rotation** of the whole em-box — correct only where the envelope edges are flat. It had **no shear term** (renderer.ts:1521 rotate → 1522 scale; the paired-edge branch returned only `{x,y,angle,vScale}`).

The envelope map between the two boundary curves `T(u)` (top) and `B(u)` (bottom) is

```
P(u, v) = (1-v)·T(u) + v·B(u),   u in [0,1] across width,  v in [0,1] top->bottom
```

Its local Jacobian at a glyph has two columns:

```
dP/du = (1-v)·T'(u) + v·B'(u)      <- advance axis, tracks the wave slope
dP/dv = B(u) - T(u)                <- em-height axis, the edge GAP vector
```

On a sloped edge these columns are **not orthogonal**. `rotate(angle)·scale` forces the vertical column perpendicular to the tangent, so it **tilts vertical strokes that should stay aligned with the gap**. For `textWave1` (parallel edges -> gap purely vertical) the mean tangent reaches **17.4 deg at the mid-band and +/-32 deg at the ends**, so the whole glyph rotated up to 17-32 deg instead of shearing.

## Fix

Return the true Jacobian. Decompose the gap vector `dP/dv = B(u)-T(u)` in the rotated (advance-axis) frame:

```
gapAlong = ( cos0·gapx + sin0·gapy) / boxHeight
gapPerp  = (-sin0·gapx + cos0·gapy) / boxHeight
vScale = gapPerp                 // em-height along the (post-shear) up-axis
shear  = gapAlong / gapPerp      // horizontal skew in the rotated frame
```

The renderer applies one extra horizontal skew between rotate and scale:

```
translate · rotate(angle) · [[1, shear],[0, 1]] · scale(hScale, vScale)
```

This reproduces the Jacobian columns exactly (reconstructed vertical column equals `(B-T)/boxHeight` to 1e-4). Vertical strokes now track the gap, the baseline follows the slope, and the glyph leans into a parallelogram.

**General, not wave-specific.** No preset name or empirical constant is branched on. On flat envelope points `gapAlong -> 0`, so `shear -> 0` and `vScale -> gap/boxHeight`, recovering the old behaviour byte-identically. Single-edge (arch/circle) Follow Path presets return `shear:0` explicitly and are untouched (they keep the rigid per-glyph rotation from #856).

## Pixel verification (headless skia render of sample-16, 96 dpi, vs the PDF)

**Wave One glyph lean** - median lean cancels on a symmetric wave, so the distinguishing metric is the *spread* of near-vertical stroke lean (structure-tensor edge orientation of anti-aliased ink):

| crop | lean IQR | p90 (forward lean) | frac lean>8deg |
|---|---|---|---|
| BEFORE (rigid rotate) | 23.4 | +23.7 | 57% |
| **AFTER (shear)** | **42.4** | **+41.4** | **85%** |
| PowerPoint PDF (truth) | 54.6 | +43.0 | 81% |

AFTER moves the stroke-lean distribution decisively from BEFORE toward the PDF. IoU(Wave1, PDF) rose 0.016 -> 0.038 (+2.4x); the low absolute IoU reflects a pre-existing vertical positioning offset, not shear quality.

**Per-preset before->after diff (slide 1, all 8 presets):**

| preset | family | before->after diff | note |
|---|---|---|---|
| textArchUp | single-edge (Follow Path) | 0.00% | unchanged |
| textArchDown | single-edge (Follow Path) | glyphs 0.000%* | unchanged |
| textCircle | single-edge (Follow Path) | 0.00% | unchanged |
| **textWave1** | **paired-edge** | **40.9%** | the fix |
| textInflate | paired-edge | 0.00% | flat at glyph points |
| textDeflate | paired-edge | 0.00% | flat at glyph points |
| textCascadeUp | paired-edge | 7.1% | subtle shear, no visual regression |
| textChevron | paired-edge | 11.7% | subtle shear, no visual regression |

*ArchDown's cell shows 3.08% total, but that is entirely Wave One's now-taller sheared ink bleeding up from the cell below - the ArchDown **glyph region (top of the cell) is 0.000% diff**. Follow Path presets are genuinely unchanged.

## Gates

- `packages/core/src/shape/text-warp.test.ts` - 19 pass (15 pre-existing byte-identical + 4 new: flat->shear 0; wave shears at slopes & reverses sign at the far end; reconstructed vertical column == gap vector; vertical strokes stay vertical on a parallel-edge wave). The single-edge arch test now also locks `shear === 0`.
- `packages/node/src/pptx-text-warp.probe.test.ts` - 4 pass. The #856 invariants (textArchUp Follow Path, textPlain flat, textInflate stretch) are unchanged; the new textWave1 shear probe **fails on the pre-shear renderer** and passes now.
- `tsc --build` (all packages) - clean.
- pptx vitest 349/349, core/shape vitest 182/182, node probes 61 pass / 11 skip (skips are demo-fixture-gated, not warp).
- pptx **demo VRT byte-identical** (all `demo/sample-1` pass; `private/sample-*` 404 because those fixtures are gitignored/absent - no warp fixture is in the VRT spec, so demo-only is sufficient per repo policy).
- No reference images touched.

## Cross-package check

docx and xlsx have **no WordArt warp draw path** (`renderWarpedText`/`buildWarpEnvelope`/`textWarp` grep = pptx + core only). The shear lives in the shared `core` geometry, so any future format that adds a warp path inherits it. No docx/xlsx changes needed.

### DIFF
--- a/packages/core/src/shape/text-warp.test.ts
+++ b/packages/core/src/shape/text-warp.test.ts
@@ -111,8 +111,12 @@ describe('textArchUp — single arc baseline', () => {
     const gl = warpGlyphTransform(env, 0.5, boxH, 0.8);
     const gr = warpGlyphTransform(env, 0.85, boxH, 0.8);
     const glft = warpGlyphTransform(env, 0.15, boxH, 0.8);
-    // Single-edge presets never compress the glyph height.
+    // Single-edge presets never compress the glyph height…
     expect(gl.vScale).toBe(1);
+    // …and never shear: Follow Path glyphs rigidly rotate along the arc.
+    expect(gl.shear).toBe(0);
+    expect(gr.shear).toBe(0);
+    expect(glft.shear).toBe(0);
     // At the apex the axis is horizontal; toward the right end it tilts down
     // (positive angle), toward the left end it tilts up (negative angle), and
     // the tilt grows monotonically away from the apex.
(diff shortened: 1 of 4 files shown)


yukiyokotani:
Reviewer nit (non-blocking, for a future doc pass): the WarpGlyphTransform JSDoc at text-warp.ts:340/345/364 narrates `hScale` as if it were a field of the transform; it is a renderer-local variable (renderer.ts:1489). The interface correctly exposes only x, y, angle, vScale, shear. Worth a one-line clarification that hScale lives in the caller.

PR #867, opened 2026-07-06
fix(docx): defer page-anchored floating table when its band intersects another float table

yukiyokotani:
## Problem

On sample-28 (an Arabic RTL competitor-qualification form), two page-anchored
`<w:tblpPr>` floating tables were drawn **stacked in the same y-range**. The
competitor-info form (§4.1) splits across pages, leaving a continuation band at
the top of the next page; the previous-projects experience form (§4.2,
`vertAnchor="page"`, `tblpY≈2174 twips = 108.7 pt`) was then painted at its raw
absolute `tblpY` **on top of** that band — an unreadable overlap of two floating
tables. As a knock-on, `renderCell` shared the page float list into each cell, so
a `vAlign="center"` header cell's text was displaced ~17 px by the overlapping
float band.

## Ground truth (Word-exported PDF, `sample-28.pdf`, 23 pages)

`pdftoppm -r 72` + visual/text (`pdftotext -layout`) inspection of pp.15–21:

| Page | Word PDF (ground truth) |
|---|---|
| p15 | competitor form (§4.1) slice 1 |
| p16 | competitor form residue **only** (company-rep info §1.4) |
| p17 | §4.2 heading + intro + previous-projects table slice 1 |
| p18 | previous-projects table residue |
| p19 | §4.3 heading + current-projects table slice 1 |
| p20 | current-projects table residue only |
| p21 | §4.4 admin-staff form (heading + block table) |

The key fact: on p16 the competitor residue sits **alone**; the projects table
begins on a **fresh** page (p17), never stacked over the residue band.

## §17.4.56 note

`<w:tblOverlap>`'s nominal default is `"overlap"` (a floating table *may* overlap
other floats). But Word's **actual** layout of a page/margin-anchored table whose
absolute `tblpY` band would land on top of another floating table already placed
on the page is to **defer the whole table to the next page**, where its identical
absolute `tblpY` no longer collides. This PR follows the measured Word behaviour,
not the spec-literal "overlap permitted" default. No heuristic constants are
introduced — the decision is a pure band-intersection test against the page's
existing table floats.

## Fix (commit 1 — primary)

`renderer.ts` `computePages`, floating-table branch (the page/margin-anchored
`first = measureFloat()` site, ~L2865): before placing slice 1, test the table's
raw (un-clamped) box against the exclusion rects of the table floats already on
the current page (`kind==='table'`, shared `FLOAT_OVERLAP_EPS` slack). On a hit —
and only when the current page is non-empty — `advancePage()` first and
re-measure. This mirrors the existing "not even the first row fits ⇒ advancePage
first" guard inside `splitFloatTableAcrossPages`. A fresh page clears page-scoped
floats, so a collision pushes the table forward exactly one page and never loops.

TDD: `float-table-page-fit.test.ts` case **(g)** — two page-anchored floating
tables whose raw bands collide ⇒ the second defers to a later page as one
un-split element; no page holds two overlapping table-float bands. Red before /
green after.

## Fix (commit 2 — secondary)

`renderCell` spread `state.floats` into `cellState` by reference, so cell
paragraphs skipped past an outer float's wrap band. Per §17.4.57 / §20.4.2.x a
table cell is its own text container (page floats exclude main-story text, not
cell text). `cellState` now gets an isolated `floats: []` / `floatParaSeq: 0`.
Verified via `onTextRun`: the sample-28 p.17 `vAlign="center"` header cell text
moves from y=282 back to its centred slot **y=264.8** (−17.2 px).

## Before / after page map (sample-28)

| | recorded ref (buggy) | this PR | Word PDF |
|---|---|---|---|
| total pages | 21 | 22 | 23 |
| p16 | competitor residue **+ projects table stacked (overlap)** | competitor residue **only** | competitor residue only |
| p17 | (projects rows overprinted) | intro + projects table slice 1 | §4.2 heading + intro + projects slice 1 |

The overlap is resolved (competitor residue and projects table no longer stacked).
The remaining 22↦23 gap is a **separate** keep-with-anchor concern: Word keeps a
section heading (e.g. the §4.2 `TOC2` heading — which has **no** `keepNext`, so
this is floating-table-anchor retention, not `keepNext`) on the same page as its
deferred table, and a second, **pre-existing** case keeps the §4.4 *block* table
with its heading. Both are orthogonal to this band-collision fix and are called
out as follow-ups.

## Verification

- **docx vitest**: 875/875 pass (incl. new case (g)).
- **root typecheck** (`tsc --build` + markdown/node/vscode): clean, exit 0.
- **ast-grep** (`ast-grep scan`): no new violations.
- **docx VRT**: sample-28 pages 1–15 are **100.0 % byte-identical**; pages 16–21
  diff **by design** (deferral re-flows the tail). The recorded reference
  (`pageCount: 21`) is **intentionally NOT updated**. All other samples
  (private 1–5, 24–27, 29–31; demo/sample-1) pass byte-identical / within
  threshold. sample-24 (floating tables) = 0-px diff.

### DIFF
--- a/packages/docx/src/float-table-page-fit.test.ts
+++ b/packages/docx/src/float-table-page-fit.test.ts
@@ -456,4 +456,61 @@ describe('computePages — floating-table page-fit / row-split (§17.4.57, Word 
       if (el) expect((el as unknown as DocTable).tblpPr).toBeDefined();
     }
   });
+
+  it('(g) DEFERS a page-anchored floating table to the next page when its raw band intersects another table float already on the page (sample-28 projects-form shape)', () => {
+    // ── Two page-anchored floating tables whose raw bands collide (§17.4.56 / Word
+    //    ground truth, sample-28 pp.16→17) ──────────────────────────────────────
+    // Content band is 100pt (pageH 140 − margins 20+20), bodyTop 20.
+    //
+    // Table A: vertAnchor="page", tblpY=10, 4 rows × 30pt = 120pt > 100pt ⇒ it
+    //   ROW-SPLITS (pageAnchoredOverflows). Slice 1 (A1–A3) fills page 1 from its
+    //   absolute page-y 10; the continuation A4 flows onto page 2 at the body top
+    //   (content-relative y 0 → 30).
+    // Table B: vertAnchor="page", tblpY=25, 3 rows × 30pt = 90pt ≤ 100pt ⇒ B alone
+    //   FITS the text region (single element, no split). Its raw absolute box is at
+    //   page-y 25 (content-relative y 5), height 90 ⇒ it would occupy content-y
+    //   5 → 95 on whatever page it lands on.
+    //
+    // If B is placed on page 2 (where A4's continuation band sits at content-y
+    // 0 → 30), B's raw band (content-y 5 → 95) INTERSECTS A4's band — two floating
+    // tables overlapping. Word's PDF (sample-28: the previous-projects experience
+    // table lands on a FRESH page after the competitor-form residue, never stacked
+    // over it) defers the whole of B to the next page, where its absolute tblpY=25
+    // no longer collides with any other table float.
+    const body = [
+      para({ text: 'a' }),
+      floatTableRows(tblp({ vertAnchor: 'page', tblpY: 10 }), 4, 30), // Table A → r1..r4
+      para({ text: 'mid' }),
+      floatTableRows(tblp({ vertAnchor: 'page', tblpY: 25 }), 3, 30), // Table B → r1..r3
+      para({ text: 'end' }),
+    ];
+    const pages = computePages(body, section(), makeCtx());
+
+    // Page 2 (index 1) carries A's continuation (A4) but MUST NOT also carry any of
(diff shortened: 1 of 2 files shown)


yukiyokotani:
Reviewer findings (non-blocking, recorded for future readers):

**F1 — evidence coupling**: with commit 1 (deferral) present, reverting commit 2 (cell float isolation) leaves sample-28 byte-identical, because the deferral removes the very overlap that displaced the header cell. Commit 2's fixture evidence (281.8→264.8, −17.0px) is only observable on the pre-commit-1 layout. Both commits were independently verified as load-bearing; commit 2 guards the general "page float overlaps a fitting cell" class per §17.4.57.

**F2 — pre-existing limitation, strictly improved but not fully solved**: in a multi-column section, two colliding page-anchored tables still stack — `nextColumnOrPage` may advance to the next column (same page, same page-scoped floats), and the collision check is single-shot (no re-loop). rawBox.y is column-independent, so the next column still collides. Not a regression (all cases stacked before this PR; single-column is now fixed), terminates safely, no fixture exercises it. Revisit if page-anchored floats in multi-column sections become relevant.

PR #872, opened 2026-07-06
feat(pptx): piecewise-affine strip subdivision for envelope-warped WordArt glyphs

yukiyokotani:
## Problem

PR #866 draws WordArt paired-edge (envelope) warps per glyph with a **single affine** — the local Jacobian of the envelope map `P(u,v) = (1−v)·T(u) + v·B(u)` sampled at the glyph centre (rotate + shear + non-uniform vertical scale). That is exact only where the map is locally affine.

**Inflate/Deflate** are the reported failure case: their top and bottom edges are mirror-symmetric, so the gap vector `B(u)−T(u)` is always vertical (shear 0, angle 0) — but its **magnitude** varies across a glyph's own width (tallest toward the box centre). A single per-glyph affine applies one uniform vertical scale, so a wide glyph came out the same height on its left and right edges, and adjacent glyphs stair-stepped instead of tracing one continuous curve. PowerPoint maps the glyph **outline** continuously through the envelope (ECMA-376 §20.1.9.19) — a fundamental limitation of one affine per glyph, not a tuning issue.

## Approach — piecewise-affine strips with deviation-driven adaptive count

**Commit 1** subdivides each glyph into vertical **strips**, each mapped by `warpGlyphTransform` sampled at **its own** centre-u; each glyph is rendered once to a reusable device-resolution offscreen and blitted per strip. As the strip count grows this converges to the true map — no preset branch, no empirical constant.

**Commit 2** (review fix) makes the strip count **adaptive**, driven by the exact error of the piecewise map. For each strip edge `b`, the true map places the glyph ruling at `Q(b,y) = A(u_b)·(0,y)` (the frame `warpGlyphTransform` returns at the boundary), while the strip paints it at `S(b,y) = A(u_centre)·(b−centre,y)`. `|S−Q|` at the slab corners is the **complete** per-strip error — it simultaneously contains:

- rotational fan-out `d·Δangle` (dominant on ring/pour envelopes, whose edge tangent sweeps up to ~345°),
- vertical-scale slide `|y|·ΔvScale` (dominant on textButton, whose mean-tangent angle is constant 0 — a pure angle-change criterion would miss it entirely),
- the baseline chord error.

Strips double until this deviation ≤ **1 device px**: two adjacent strips each within 1 px of the *same* true ruling diverge ≤ 2 px, which the 2 device px of shared overlap provably covers — that is the derivation tying the budget to the overlap. If a doubling fails to reduce the deviation by ≥ 25 %, the residual sits at a C0 corner of the envelope itself (chevron apex — a genuine crease that refinement can only localize), so the loop keeps the finer partition and stops instead of spinning to the `WARP_STRIP_MAX_PER_GLYPH = 256` cap. **No preset fallback branch anywhere.**

**Single-edge (Follow Path — Arch/Circle, #856) is untouched and byte-identical.** Device scale is read lazily on the first paired-edge glyph only.

## Error analysis (corrected per independent review)

The original claim "8-px strips keep error < 0.3 px for every preset" was **wrong** — it held only near the baseline of the low-curvature fixture presets. Corrected measurements for the fixed 8-device-px width (before the adaptive loop): baseline-vicinity Inflate 0.72 / Wave1 0.19 / Chevron 0.12 px, but at em-box extremes Wave1 ≈2.4 / Button ≈5.9 / CirclePour ≈11 / CanUp ≈44 px — the ring/pour families shattered into radial slivers (a 5-letter word: **156 ink components** on textRingOutside, 21 on textCirclePour).

With the adaptive loop (representative 28-css-px glyph, dpr 2):

| Preset | k (width → adaptive) | deviation (device px) |
|---|---|---|
| textButton | 30 → 240 | 6.07 → **0.76** |
| textCirclePour | 30 → 240 | 5.61 → **0.70** |
| textRingOutside | 30 → 240 | 4.73 → **0.59** |
| textArchUpPour | 30 → 60 | 1.59 → **0.79** |
| textInflate / Deflate / Wave1 / Cascade / Chevron | 30 → 30 | 0.03–0.71 (already ≤ budget) |

Ink components ("Round"): CirclePour 21 → **5**, RingOutside 156 → **4**, CanUp/Button/ArchUpPour stable at 5. The fixed renders are contiguous, readable text bending around the ring — better than both the sliver regression *and* the single-affine baseline (rigid flat letters at discrete rotations).

### Intra-glyph asymmetry (the originally reported defect) — unchanged by the review fix

Wide glyph in the box's left half; right (centre-facing) edge ink-height ÷ left edge:

| Preset | single affine | strips | expected (PowerPoint PDF) |
|---|---|---|---|
| Inflate | 0.98 | **1.30** | centre bulge ✓ |
| Deflate | 0.98 | **0.72** | centre pinch ✓ |

Regression test asserts `right/left > 1.12` for Inflate; fails on the single-affine baseline (0.977).

### Reviewed presets preserved after the adaptive change

Pixel diff of the adaptive build vs the previously reviewed strip build: Inflate / Cascade / Chevron / ArchUp / ArchDown / Circle **0.000 %**; Deflate 1.13 % and Wave1 0.77 % refine further toward the exact map (those glyphs exceeded the 1 px budget and subdivided). vs the single-affine baseline the paired-edge presets still differ 5–7.6 %, single-edge 0.000 %.

### Seams

Hairline-gap scan: 0 candidates on all fixture presets; 1–2 isolated 2-px candidates on CirclePour/ArchUpPour whole-word renders (vs thousands of gap pixels in the sliver regression).

### Performance

`renderSlide`, dpr 2, median: fixture slide (8 warps) 2.5 ms (single affine) → 10.6 ms (fixed width) → **11.9 ms** (adaptive). Worst high-curvature case: textCirclePour, 5 glyphs × k=240 ≈ 1200 blits → **6.8 ms**/slide; textRingOutside 5.8 ms.

### Gates

- pptx vitest **349 passed**, core shape **182 passed**, core text-warp **19 passed**
- node warp probe **6 passed** (+ sliver guard: components ≤ 8 on CirclePour/RingOutside/CanUp — fails on the width-only revision with 34 components, passes here)
- pptx renderer + node probe typecheck **0 errors**; ast-grep unchanged (3 pre-existing warnings, no new `as unknown as`)
- pptx demo VRT: all demo tests pass — visual regression 9 slides byte-identical, worker equivalence (3 slides + selection/findText/zoom + IX-nav) ✓

## Files
- `packages/pptx/src/renderer.ts` — strip subdivision draw path + deviation-driven adaptive strip count (paired-edge only)
- `packages/node/src/pptx-text-warp.probe.test.ts` — factory to exercise strips in CI, intra-glyph asymmetry regression test, high-curvature sliver guard

### DIFF
--- a/packages/node/src/pptx-text-warp.probe.test.ts
+++ b/packages/node/src/pptx-text-warp.probe.test.ts
@@ -421,4 +421,63 @@ describe.skipIf(!skia)('node WordArt text-warp geometry (prstTxWarp)', () => {
     // Measured: single-affine ≈1.00 (fails), strips ≈1.30 (passes).
     expect(rightH / leftH).toBeGreaterThan(1.12);
   });
+
+  it('high-curvature envelopes (rings/pours) stay contiguous — no radial slivers', async () => {
+    // Regression guard for the ADAPTIVE strip count. On ring/pour envelopes the
+    // edge tangent sweeps up to ~345° across the width, so with a fixed
+    // width-only strip count adjacent strips rotate apart faster than the ±1px
+    // overlap can bridge: each glyph shatters into a fan of disconnected radial
+    // slivers (measured: "WARP" in textRingOutside split into dozens of
+    // 8-connected ink components; a contiguous render has one component per
+    // detached glyph part). The deviation-driven subdivision keeps every strip
+    // within 1 device px of the exact envelope map, so the ink stays connected.
+    // "WARP" is four single-piece letters — allow a small margin for warped
+    // letters splitting at hairline joins, but a sliver fan (>>8) must fail.
+    for (const preset of ['textCirclePour', 'textRingOutside', 'textCanUp']) {
+      const width = 400;
+      const height = 240;
+      const canvas = new Canvas(width, height);
+      await renderSlideNode(
+        canvas as unknown as Parameters<typeof renderSlideNode>[0],
+        buildSlide(preset),
+        0,
+        { width, dpr: 1, factory: warpFactory },
+      );
+      const ctx = canvas.getContext('2d');
+      const data = ctx.getImageData(0, 0, width, height).data as unknown as Uint8ClampedArray;
+      const grid = new Uint8Array(width * height);
+      let inkPx = 0;
+      for (let p = 0; p < width * height; p++) {
+        const i = p * 4;
+        const lum = 0.299 * data[i] + 0.587 * data[i + 1] + 0.114 * data[i + 2];
+        if (data[i + 3] > 40 && lum < 128) { grid[p] = 1; inkPx++; }
+      }
+      expect(inkPx).toBeGreaterThan(100);
+      // 8-connected component count over the ink mask.
+      const seen = new Uint8Array(width * height);
+      const stack: number[] = [];
+      let comps = 0;
+      for (let start = 0; start < width * height; start++) {
+        if (!grid[start] || seen[start]) continue;
+        comps++;
+        stack.length = 0;
+        stack.push(start);
+        seen[start] = 1;
+        while (stack.length) {
+          const p = stack.pop() as number;
+          const x = p % width, y = (p / width) | 0;
+          for (let dy = -1; dy <= 1; dy++) for (let dx = -1; dx <= 1; dx++) {
+            if (!dx && !dy) continue;
+            const nx = x + dx, ny = y + dy;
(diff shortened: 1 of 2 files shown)


yukiyokotani:
Re-review APPROVE (adaptive subdivision @ eb023d2). Reviewer nit for a future doc pass: the cap-256 comment says the residual deviation "is reported by the loop's last measurement" — `dev` is only the loop-condition variable and is not surfaced externally; "measured" would be more accurate than "reported".

PR #915, opened 2026-07-10
refactor(xlsx): consolidate image cache onto shared getCachedBitmapByPath

### LINKED ISSUE #781: xlsx: consolidate image cache onto shared getCachedBitmapByPath (post-#779)
## Summary
Follow-up from the #779 review (accurate scope, correcting the commit's overstated deferral rationale). #779 fixed the xlsx ImageBitmap teardown leak with a local close-loop. The cleaner long-term shape is to consolidate xlsx onto the shared `getCachedBitmapByPath`/`dropBitmapCacheByPath` cache that docx/pptx already use (CLAUDE.md cross-package principle).

Actual remaining work (verified in review, not a "sweep"):
- `render-orchestrator.ts` `prefetchImages`/`decodeImageSource`: replace the hand-rolled `Map` and split the mixed raster/SVG cache the way docx/pptx do.
- Only **2** synchronous lookup sites in `renderer.ts` (3187, 3426) to repoint.

## Also (pre-existing, cross-package, low severity)
None of docx/pptx/xlsx render-workers explicitly evict the shared worker-mode `ImageBitmap` cache on re-parse (only LRU-bounded). Symmetric across all three (same "single-document contract" comment). Not a leak (bounded), but worth an explicit drop on re-parse for consistency. Surfaced in the #779 review.

yukiyokotani:
## Summary

Follow-up to the #779 review. #779 fixed the xlsx `ImageBitmap` teardown
leak with a local close-loop (`closeAndClearImageCache`); this consolidates
xlsx onto the shared, per-`fetchImage` core caches that docx and pptx
already use, then adds the symmetric re-parse teardown across all three
render workers. Closes #781.

## Design (shared-API contract, no xlsx-only branch)

xlsx previously decoded blips into a plain `Map<string, CanvasImageSource | null>`
that **owned** the resulting `ImageBitmap`s and needed a bespoke
`closeAndClearImageCache` to `.close()` them on destroy / re-parse. docx and
pptx instead decode through core's shared caches, which own the GPU-backed
bitmaps, bound them with an LRU, close them on eviction, and release them
per document via matching `drop*` calls:

- base raster / metafile → `getCachedBitmapByPath`
- `<a:duotone>` recolour (§20.1.8.23) → `getCachedDuotoneBitmapByPath`
- SVG vector original → `getCachedSvgImageByPath`

Changes:

- **`render-orchestrator.ts`** — `decodeImageSource` now routes the raster
  path through `getCachedDuotoneBitmapByPath` (a thin two-layer wrapper over
  `getCachedBitmapByPath`), folding the former standalone `applyDuotone`
  step into the shared decode; the SVG branch stays on
  `getCachedSvgImageByPath`. `prefetchImages` re-resolves every referenced
  image against the shared caches each pass — the way docx/pptx do — so a
  still-referenced blip whose bitmap was LRU-evicted (and closed) is
  transparently re-decoded rather than served closed. A shared-cache hit
  re-fetches no bytes and re-runs no decode.
- **The per-instance / per-worker `imageCache` map** becomes a pure
  synchronous-lookup layer for the grid draw, keyed by `imageCacheKey`,
  exactly like docx's `preloadImages` map. It no longer owns the bitmaps,
  so **the two synchronous lookup sites in `renderer.ts` need no change** —
  they still read that map by key. (The issue's "repoint 2 sites" assumed a
  peek-into-shared-cache design, but xlsx's fully-synchronous grid draw
  needs a mixed-type lookup covering raster + SVG + duotone, which
  `peekCachedBitmapByPath` doesn't cover; keeping the lookup map is docx's
  exact idiom for a synchronous draw.)
- **`closeAndClearImageCache` removed.** `XlsxWorkbook.destroy()` and the
  render worker's re-parse clear the lookup map and drop the three shared
  caches keyed by their `fetchImage` closure (`dropBitmapCacheByPath`,
  `dropDuotoneBitmapCache`, `dropSvgImageCache`).

No shared-API branch was added for xlsx: `getCachedDuotoneBitmapByPath`
already existed in core (used by pptx), and its own doc-comment noted xlsx
had not yet adopted it. xlsx's `<a:srcRect>` full-frame metafile sizing
stays in the xlsx caller as the `metafileRasterSize` opts it feeds the
shared decode — the same idiom pptx uses.

## Secondary (cross-package, separate commit)

None of the docx/pptx/xlsx render workers evicted the shared worker-mode
decoded caches on re-parse (byte caches only). Because each worker's
`getImage` closure is a stable module-level identity, a reused worker
re-parsing a document with an identically-named zip path could be served
the previous file's decoded bitmap, and the GPU / object-URL handles
lingered until the LRU cap. All three workers now drop the same caches
their `destroy()` drops, keyed by `getImage`.

## Eviction semantics pinned by tests (no #779 regression)

- After `prefetchImages`, dropping the shared **base** / **duotone** caches
  closes the decoded bitmaps (proving ownership moved to the shared cache —
  no teardown leak); clearing the lookup map never closes (non-owning, no
  double close).
- `XlsxWorkbook.destroy()` closes a base bitmap and a duotone recolour that
  were decoded into the shared caches keyed by the instance's `_fetchImage`,
  and empties the lookup map; safe to call twice.
- A second `prefetchImages` pass re-fetches nothing (shared-cache hit).

## Verification

- `vitest run packages/xlsx` — 510 passed.
- `vitest run packages/docx packages/pptx` — 1347 passed, 1 pre-existing skip.
- Node device-pixel probe `xlsx-blip-duotone-alpha.probe.test.ts` (renders
  a duotone + `alphaModFix` blip through the refactored orchestrator with
  skia) passes unchanged — pixel behaviour is identical, as expected for a
  pure refactor.
- Root `pnpm typecheck` clean.

### DIFF
--- a/packages/core/src/image/bitmap-image-by-path.test.ts
+++ b/packages/core/src/image/bitmap-image-by-path.test.ts
@@ -3,6 +3,7 @@ import {
   getCachedBitmapByPath,
   peekCachedBitmapByPath,
   dropBitmapCacheByPath,
+  acquireBitmapCacheLease,
 } from './bitmap-image-by-path';
 
 /** Build a minimal standard (non-placeable) WMF that draws one polyline, so the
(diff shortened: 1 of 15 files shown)


yukiyokotani:
Addressed the independent review's two blocking findings in 88e7429:

**1. LRU eviction closing bitmaps an in-flight pass still references (high).** Fixed structurally in the shared layer: core now exposes `acquireBitmapCacheLease(fetchImage)`. While a render pass holds a lease, every close the per-document caches would perform — LRU eviction past the cap, `dropBitmapCacheByPath`, `dropDuotoneBitmapCache`, docx's `dropColorReplacedCache` — is deferred to the last release (eviction still removes the entry immediately, so the cache stays size-bounded; the next resolve re-decodes). No cap changes, no magic numbers. Investigation confirmed the same hole in docx (`preloadImages` map → synchronous page paint) and a narrower variant in pptx (eviction between the draw loop's await and its `drawImage` under warm-pass pressure), so the lease is wired around all three render passes: xlsx `renderWorksheetViewport`, docx `renderDocumentToCanvas`, pptx `renderSlide` (thin wrappers around the verbatim bodies).

**2. Stale lookup entry surviving a failed re-resolve (high).** `prefetchImages` now deletes the key in its catch: the entry is re-resolved precisely because its shared-cache backing may be evicted+closed, and the renderer skips only a missing/falsy source.

Tests: 5 new core lease-semantics tests (over-cap eviction deferred, drop-under-lease deferred with immediate cache forget, nested leases, idempotent release, unleased baseline); an end-to-end xlsx probe rendering 300 images (cap 256) through `renderWorksheetViewport` asserting `drawImage` never receives a closed bitmap and exactly N−cap deferred closes run after the pass — this probe reproduced the closed-at-draw failure before the wiring (Red) and passes after (Green); a failure-path deletion test. The three shared-cache ownership tests are unchanged. Full core/xlsx/docx/pptx suites (3376 passed), root typecheck, ast-grep lint, and the skia duotone+alpha device-pixel probe all pass.

yukiyokotani:
Addressed the lease-mechanism re-review's three findings in 775f4ec:

**1. Derived-cache closed-bitmap hole (high).** A pass-through second-layer entry (recolour resolved to the base bitmap itself — pixel pipeline unavailable / degenerate size) could outlive the base entry: after base LRU eviction closed the base, the stale second-layer hit bypassed the base layer's remove-on-evict → re-decode protection and served the closed bitmap to the next pass. Pass-through entries now self-evict on resolution (`bmp === base`), so the next pass re-resolves through the base cache and always sees a live bitmap; fresh recolour rasters (exclusively owned) stay memoized. Applied identically to core `getCachedDuotoneBitmapByPath` and docx `decodeRaster` (clrChange layer). Red verified by temporarily disabling the self-evict.

**2. xlsx fetchImage identity (medium).** `XlsxWorkbook.renderViewport` now always renders under the instance's stable `_fetchImage` closure (caller-supplied `opts.fetchImage` no longer overrides, matching docx/pptx) — that identity keys the shared caches, the lease counter, and `destroy()`'s drops, so a per-call closure would split the namespace and leave bitmaps undropped at destroy. Documented on the method; external callers needing a custom byte source use the standalone orchestrator (self-consistent per-call identity including its lease). No in-repo caller passed `fetchImage` to the workbook method.

**3. Double close (low).** All GPU closes in the shared caches funnel through a WeakSet-deduplicated `closeBitmapOnce`, removing the reliance on `ImageBitmap.close()` idempotence when two layers resolve to the same bitmap.

Tests: core pass-through-never-outlives-base + exactly-once-close-on-double-drop; xlsx renderer-level interaction (duotone pass-through replaced with a live bitmap on the pass after base eviction) + `renderViewport` identity pinning. Full core/xlsx/docx/pptx suites (3380 passed), root typecheck, ast-grep lint, and the skia duotone+alpha pixel probe all pass. Implementation reviewed and the two xlsx tests authored by Codex (gpt-5.6-sol); both co-author trailers included.

PR #918, opened 2026-07-10
feat(docx): implement w:fitText manual run width (ECMA-376 §17.3.2.14)

yukiyokotani:
## Summary

Implements ECMA-376 §17.3.2.14 `<w:fitText>` (Manual Run Width) for the DOCX renderer: a run's contents are displayed within the width given by `w:val` (twips), and **consecutive runs carrying the same `w:id` merge into one fit region** that fills `w:val` together (the spec's example: a 1-inch region split across three runs occupies one inch in total). An id-less fitText run never links to a neighbour, and a non-fitText run breaks adjacency.

Previously the element was ignored entirely, so fitted labels (common in Japanese government/business forms, where a column of headings is aligned to one fixed width) rendered at their natural width plus whatever stale `w:spacing` Word had cached — misaligning the whole label column.

## Geometry

Measured from a private fixture's Word-generated PDF (the acceptance ground truth):

- Glyphs keep their **natural advance** (including the §17.3.2.43 `w:w` horizontal glyph scale).
- The slack `(val − Σnatural)` is distributed **evenly as inter-character gaps** — one gap between every adjacent glyph pair across run boundaries, **no gap after the region's last glyph**:

  ```
  perGap = (targetPx − naturalPx) / (charCount − 1)
  ```

- Cached §17.3.2.35 `w:spacing` on the runs is **ignored** inside a fit region: Word recomputes the pitch from `w:val`, and the fixture's cached value demonstrably disagrees with its own PDF.
- Compression (`Σnatural > val`) uses the same formula with a negative gap. The spec phrases compression as "decreasing the size of each character"; this is recorded in-code so a charScale-based approach can be swapped in if compression ground truth appears.
- A fit region is an **atomic, non-wrapping unit** (a fixed-width cell of width `val`).

## Implementation

- **Parser (Rust)**: `<w:fitText w:val w:id>` → `TextRun.fitTextVal` (kept in twips so the layout layer applies its px-per-pt scale exactly once) + `fitTextId` (signed ST_DecimalNumber). Follows the shared run-property cascade (set overrides, absent inherits) beside the §17.3.2.35/43/24/19 axes.
- **Kernel (`fit-text.ts`)**: pure `groupFitTextRegions(runs, scale)` — the single source of region grouping and per-gap math, pinned by 12 standalone unit tests (merge, adjacency break, id-less standalone, twips→px scale, `w:w` composition, negative gap, single-glyph, no-fitText).
- **Layout (`line-layout.ts`)**: regions are grouped over **source runs** in `buildSegments` (script/small-caps segmentation cannot fork a region), resolved from raw Canvas advances at the exact layout scale, and folded into the existing advance authority (`segAdvanceWidth` / `segLetterSpacingPx` / `charSpacingDeltaPx`) — no parallel width system. Non-last region segments carry one trailing boundary gap; the last carries none, so the region sums to `val` exactly. The line breaker treats the region as one glued, atomically-fitted unit, and `rescaleLayoutLines` re-resolves the gap from paint-scale metrics so measure == paint.
- **Paint (`renderer.ts`)**: a fit segment draws through the same uniform `ctx.letterSpacing` model as the §17.18.44 fully-distributed justify arm, composed with `w:w` exactly like the sibling arms (fixed pitch divided by the glyph scale inside the ×scale frame). Paragraph justification does not re-stretch a fit region's internal gaps. Find/selection coordinates (`onTextRun`) and run decorations follow the fitted advance automatically via `measuredWidth`.

## Verification

- fitText suites: 17/17 (12 pinned kernel + layout integration + recording-canvas paint).
- Full docx suite: 141 files, 1031 passed, 1 skipped — no regressions.
- Root `pnpm typecheck`: pass. Rust: `cargo fmt --check`, `clippy -D warnings`, 312 tests pass. WASM rebuilt via the package script (reinit export verified).
- Node probe (skia) against the private fixture's Word PDF, three fit regions (`val=2400` = 120 pt):

  | Region | before advance (pt) | after advance (pt) | Word PDF |
  |---|---|---|---|
  | 6 glyphs, 3 runs | 96.00 | **120.00** | 120.00 |
  | 4 glyphs, 3 runs | 84.00 | **120.00** | 120.00 |
  | 15 glyphs, `w:w`=66% | 119.70 | **120.00** | 120.00 |

  Head-run advance 86.40 pt (= 48 + 4×9.6, the recomputed per-gap — not the cached `w:spacing`), region right edge within 0.06 pt of the PDF.

### DIFF
--- a/packages/docx/parser/src/numbering.rs
+++ b/packages/docx/parser/src/numbering.rs
@@ -7,30 +7,17 @@ use std::collections::{HashMap, HashSet};
 
 /// Parse a single VML CSS length (e.g. `width:9pt`) from a `style` attribute
 /// into pt. Supports the units Word emits for picture-bullet shapes: `pt`
-/// (1pt), `in` (72pt), `pc` (12pt), `cm` (28.3465pt), `mm` (2.83465pt). A bare
-/// number with no unit is treated as pt (VML's default user unit for shapes is
-/// the point). Returns `None` when the property is absent or unparseable.
+/// (1pt), `in` (72pt), `pc`/`pi` (12pt), `cm` (28.3465pt), `mm` (2.83465pt). A
+/// bare number with no unit is treated as pt (VML's default user unit for
+/// shapes is the point). Returns `None` when the property is absent or
+/// unparseable.
 fn vml_style_len(style: &str, prop: &str) -> Option<f64> {
     for decl in style.split(';') {
         let (k, v) = decl.split_once(':')?;
         if k.trim() != prop {
             continue;
         }
-        let v = v.trim();
-        let (num, factor) = if let Some(n) = v.strip_suffix("pt") {
-            (n, 1.0)
-        } else if let Some(n) = v.strip_suffix("in") {
-            (n, 72.0)
-        } else if let Some(n) = v.strip_suffix("pc") {
-            (n, 12.0)
-        } else if let Some(n) = v.strip_suffix("mm") {
-            (n, 2.834_645_7)
-        } else if let Some(n) = v.strip_suffix("cm") {
-            (n, 28.346_457)
-        } else {
-            (v, 1.0)
-        };
-        return num.trim().parse::<f64>().ok().map(|n| n * factor);
+        return parse_measure_to_pt(v, 1.0);
     }
     None
 }
--- a/packages/docx/parser/src/parser.rs
+++ b/packages/docx/parser/src/parser.rs
@@ -2977,6 +2977,16 @@ fn text_runs_mergeable(a: &TextRun, b: &TextRun) -> bool {
         && a.italic_cs == b.italic_cs
         && a.lang_bidi == b.lang_bidi
         && a.snap_to_grid == b.snap_to_grid
+        // Character metrics change measured or painted geometry, so every one
+        // must match before a noBreakHyphen run can be folded into its neighbour.
+        && a.char_spacing == b.char_spacing
+        && a.fit_text_val == b.fit_text_val
+        && a.fit_text_id == b.fit_text_id
+        && a.char_scale == b.char_scale
+        && a.position == b.position
+        && a.kerning == b.kerning
+        && a.east_asian_vert == b.east_asian_vert
+        && a.east_asian_vert_compress == b.east_asian_vert_compress
 }
 
 // Same parse-context threading as handle_run_in_para.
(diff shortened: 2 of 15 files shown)


yukiyokotani:
## Adversarial review round — 8 adjudicated findings fixed

An independent adversarial review of this PR produced 9 findings; 8 were adjudicated for fixing (the RTL cross-segment gap finding was deferred — no fixture/ground truth). All 8 are now addressed in three commits, each preceded by failing (Red) contract tests:

**Layout geometry** (`69ba307`)
1. **[high] Single-character region** now occupies exactly `w:val`: the kernel computes the residual as a trailing cell pad on the region-end segment (algebraically zero for multi-glyph regions); the glyph keeps its natural width at the region start. Glyph scaling stays deferred until compression ground truth exists.
2. **[high] Tab inside a fitText run** splits the region at the tab boundary (a tab advance is not a glyph advance): each text fragment fits `w:val` independently and never links across the tab; charCount no longer counts the tab.
3. **[medium] Case transforms changing the code-point count** (ß → SS under small caps): the region charCount now aggregates from the emitted segment texts, keeping the advance pinned to `w:val`.

**Justify + overlays** (`564d451`)
4. **[medium] Justified lines containing a fit region** no longer fall short of the right margin: fit segments enter the §17.18.44 distribution as non-text objects (no internal gaps), so the slack redistributes to legal gaps.
5. **[high] Fitted geometry now reaches selection and find**: `DocxTextRunInfo.letterSpacingPx` carries the drawn per-code-point pitch (same `segLetterSpacingPx` authority as the paint); the text layer applies it as the span's CSS letter-spacing and the find-highlight layer offsets/widens slices by it. Shared core helpers untouched.

**Parser contract** (`e09d277`)
6. **[medium] Composite cascade**: `w:fitText` cascades as one unit — a direct element without `w:id` clears an inherited id.
7. **[medium] Schema domains**: `w:id` (ST_DecimalNumber, arbitrary-precision XSD integer) is preserved as a string end-to-end; `w:val` accepts positive universal measures (`1in` = 1440 twips) via a shared suffix parser also reused by the VML length path.
8. **[medium] `text_runs_mergeable`** now gates on all character-metric fields (fitText, `w:spacing`, `w:w`, `w:position`, `w:kern`, `eastAsianLayout`), so a noBreakHyphen can no longer absorb a run with different drawn geometry.

**Verification (branch head)**: fitText suites 28/28 (13 new Red-first contracts); full docx 142 files / 1042 passed; root typecheck; cargo fmt/clippy `-D warnings`/316 tests; WASM rebuilt via the package script; the private-fixture probe re-measured identically (all three regions advance 120.00 pt, right edge within 0.06 pt of the Word PDF).

PR #924, opened 2026-07-10
fix(docx): size docGrid line cells from the run em, not the glyph box

yukiyokotani:
## Problem

In an ECMA-376 §17.6.5 document-grid section (`<w:docGrid w:type="lines" …>`),
each single-spaced East Asian line snaps to a whole number of grid cells. The
cell **count** was derived from the Canvas glyph box
(`fontBoundingBoxAscent + fontBoundingBoxDescent`), which for CJK faces is
~1.602 em. So a 12 pt run on an 18 pt line pitch computed
`ceil(12 * 1.602 / 18) = 2` cells = 36 pt, while Word snaps it to **1 cell =
18 pt**. Every over-counted line pushed the following content ~18 pt down, so
body paragraphs above an anchored object drifted below their Word positions.

## Fix

The cell count is a function of the run **em** (font size), not the substituted
font's leading. This adds a pure helper and threads the line em into
`lineBoxHeight`:

```
docGridLineCells(emPx, pitchPx) = floor(emPx / pitchPx) + 1
```

### Why `floor(em/pitch) + 1`

Word PDF exports (`pdftotext -bbox`) give three data points:

| em | pitch | Word cells |
|----|-------|-----------|
| 10.5 pt | 18 pt | 1 |
| 12 pt | 18 pt | 1 |
| 20 pt | 20 pt | 2 |

`ceil(em/pitch)` fails the `em == pitch` boundary (gives 1, Word wants 2).
`floor(em/pitch) + 1` fits all three: an em square that exactly fills *k*
pitches still needs its inter-line leading and spills into cell *k+1*. The
sub-pitch region and the `em == pitch` boundary are measured; the `k ≥ 2`
region extrapolates the scale-free boundary rule (no Word measurement covers
it yet — recorded as such in the contract comment). ECMA-376 §17.6.5 defines
`linePitch` as the height of one single-spaced line; the whole-cell rounding of
taller lines is Word runtime behaviour (no empirical thresholds or ratio
constants are introduced).

### Scope

- The rule is gated **per line**, on the line's script (`EAST_ASIAN_RE` over
  its text segments) — a Latin-only line inside a CJK paragraph keeps its
  natural height above a one-cell floor, matching the text-box paths.
- The line's em is the **tallest run** on the line (a conservative
  anti-clipping choice consistent with §17.3.1.33); a manual line break's run
  size participates in the line height but no longer overrides a taller peer.
- **Ruby lines** (body and text-box) are excluded: their measured glyph box
  reserves the base text plus the raised furigana annotation, which the em
  would collapse and clip — they keep the measured-glyph-box cell rounding.
  Ruby paragraphs also keep their paragraph-uniform line height unchanged.
- Latin, off-grid, `exact` / `auto`, and non-grid paths are unchanged.
- `atLeast` on an active grid **changes**: its grid-minimum term is now the
  em-based single cell instead of the glyph-box rounding, so a 12 pt line with
  a 19.22 px glyph box on an 18 pt pitch resolves to its natural 19.22 px
  (previously 36 px). No corpus sample exercises this combination and a Word
  fixture export could not be captured, so the exact Word behaviour is
  unverified; the current resolution is pinned by dedicated `[pin]` tests so
  any future change is deliberate.

## Review follow-ups (second commit set)

An adversarial review of the first commit found three call-path regressions,
each now pinned by a failing-first unit test:

1. **Text-box ruby** — both text-box height paths hard-coded `hasRuby=false`;
   they now forward `LayoutLine.hasRuby` so §17.3.3.25 ruby lines keep their
   furigana reserve (render + `spAutoFit` measure).
2. **Manual line break** — `<w:br>` overwrote the line's tallest run size with
   the break run's own size, shrinking the em (20 pt CJK line with a 10 pt
   break run: 2 cells → 1). Fixed with `Math.max`.
3. **Per-line script gate** — body call sites used the paragraph-level East
   Asian flag; a per-line flag now feeds every non-ruby resolution.

## Verification

- `@silurus/ooxml-docx` unit suite: **1088 passed** (new cell-count, ruby,
  manual-break, per-line-gate, and atLeast-pin cases included).
- Root `tsc --build`: **0 errors**.
- Headless render (real Yu Mincho) of a line-grid East Asian document matches
  the Word baselines: the top line unchanged, the heading and following lines
  rise onto Word's grid (band deltas 36.0 / 18.0 / 18.0 pt vs Word's
  35.6 / 18.5 / 17.9).
- A line-grid document with a 20 pt title stays 2 cells (40 pt/line).
- Before/after render of the line-grid visual-regression samples: non-East-
  Asian and ruby grid content is byte-identical; a vertical (tbRl) line-grid
  sample moves toward its Word reference; **no page moves away**. Unchanged
  after the review follow-ups.

Note: the committed Playwright VRT targets Google Chrome (`channel: "chrome"`),
which is not installed in this environment, so the browser VRT itself was not
executed; the direction was verified with the headless render comparison above.

### DIFF
--- a/packages/docx/src/line-box-height.test.ts
+++ b/packages/docx/src/line-box-height.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest';
-import { isGridLineRule, lineBoxHeight } from './line-layout.js';
+import { docGridLineCells, isGridLineRule, lineBoxHeight } from './line-layout.js';
 import type { LineSpacing } from './types.js';
 
 // Regression: some generators emit <w:spacing w:line="0" w:lineRule="exact"/> on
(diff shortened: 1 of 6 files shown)


yukiyokotani:
## Adjudication: atLeast × line grid — current implementation is correct

Fixture `sample-43` (atleast-ea-grid): identical 10.5pt CJK text on an 18pt line grid (`<w:docGrid w:type="lines" w:linePitch="360"/>`), three paragraphs differing only in `w:line` (atLeast). Baseline-to-baseline pitch measured from Word's PDF export via `pdftotext -bbox` (yMin deltas of the wrapped CJK lines, exact to 0.001pt):

| Paragraph | `w:line` (atLeast) | Measured pitch | cell-multiple model | max(cell,atLeast) model |
|---|---|---|---|---|
| A | 240 (12pt, < pitch) | **18.0pt** | 18 | 18 |
| B | 360 (18pt, = pitch) | **18.0pt** | 18 | 18 |
| C | 480 (24pt, > pitch) | **24.0pt** | 36 (2 cells) | 24 |

C is the discriminator: Word yields **24pt**, i.e. it honours the raw atLeast value once it exceeds one grid cell — it does **NOT** snap up to the next whole-cell multiple (36pt). The effective pitch is `max(gridCell, atLeast, natural)` = 18/18/24, which is exactly what `max(natural, authored, single-em-cell)` produces here. **No change needed.** The "Word not yet verified" pin from this PR can be updated to these measured values.

PR #926, opened 2026-07-10
Split restart-merge rows mid-page and open the page-cut edges

yukiyokotani:
## Split rows that start a vertical merge; leave mid-row page cuts open

Stages C/D of the mid-page table-row-break series (stage B = #925). Two findings from
the target document class (a multi-page government form; described here without private
fixture details) drove this:

### 1. Rows beginning a §17.4.85 vertical merge never split (`fix(docx): split rows that start a vertical merge across the page band`)

Mid-page row splitting already ships for ordinary rows, but BOTH row splitters rejected
any vMerge involvement — so a tall form row with narrow `vMerge=restart` label columns
and a long content cell moved to the next page whole, underfilling the source page where
Word splits at a cell line boundary (§17.4.6: splittable by default).

The spec-faithful gate distinguishes the two §17.4.85 roles: a CONTINUE cell's box is
owned by an earlier row (still excluded); a RESTART cell starts its span in THIS row —
its content fits the band like any cell's, the page-1 piece keeps `restart` (truncated
box ends at the cut), the page-2 piece keeps `restart` too, so the following `continue`
rows chain onto it via `findMergeEndRow` and the span re-opens on the continuation page
exactly as Word draws it. `tableBreakAllowedBefore` keeps the re-opened span attached to
its continue rows. Pinned Red-first with a synthetic mirror of the class (both label
variants: fully-consumed → empty continuation, and label splitting alongside the content
cell), fragment provenance (`sourceRowIndex` shared, restart/continue/continue roles),
and the continue-cell refusal.

Verified against the target document (local render probe vs its Word PDF): the row now
splits mid-page with the span re-opening on page 2 — structurally matching Word. The
exact split line differs by ~2 lines and the label column's line pitch differs (accumulated
font-metric/line-spacing drift over the preceding ~750pt, pre-existing), left for
fidelity adjudication.

### 2. Mid-row page-cut edges drew row borders (`fix(docx): leave mid-row page-cut edges open`)

A page cut THROUGH a row is not a row boundary: Word leaves the page-1 piece's bottom
edge open (the row continues) and the continuation draws its own top edge — the target
document's PDF shows exactly this. Our renderer closed the cut with the piece's resolved
bottom border. Cut pieces now carry a runtime-only marker (on the row clone, never the
parsed model) consumed by the shared border pass; row-boundary cuts are untouched
(pinned). The §17.4.66 ownership convention also collapses the would-be rule between two
pieces stacked on one page.

### Review fixes (High blocking + Low, both addressed; branch rebased onto #925's tip)

- **[High] Span-extension double counting**: after a restart-row split, the pieces
  carried the fitted restart content while the merge-end row kept the §17.4.85
  extension resolved for the UNSPLIT row (the reviewer's 100pt-body repro produced a
  160pt page). The final piece's height now EXCLUDES restart remainders (they
  distribute over the re-opened span; the splitter reports them per grid column since
  the piece content is line-sliced and must not be re-measured whole), and a repair
  pass at all three splice sites resets the tail rows to base heights and re-applies
  the extension exactly once in resolver order. The slice-start trigger is now
  unbreakable-GROUP aware, so an over-tall restart+continue group splits its head into
  the remaining band and the re-opened piece re-splits on the next page when the
  remaining span still exceeds a fresh page. Red-first with the review's repro shape
  (every page ≤ the body band; every label paragraph and content line placed exactly
  once, whole or sliced).
- **[Low] vAlign coverage**: the target class centres its restart labels, so excluding
  center/bottom would regress it. Structure pinned for both label variants
  (split fires, content once, band bound); the per-piece re-centring is documented as
  Word-unverified at both splitter gates instead of asserting unknown Y positions.

Full gate re-run: docx 1134 tests green; VRT unchanged from the pre-fix branch state
(the only delta vs base remains the border commit's known-fail-page improvement; the
span repair itself is pixel-inert across all 89 checks).

### Verification

- Full workspace `pnpm test` green (docx 145 files / 1125 tests); root typecheck, lint,
  lint:test, `cargo fmt --check`, `git diff --check` all clean.
- Local pixel VRT (89 checks): ONE check moved vs the base merge commit — a known-fail
  page improved by 478 diff px (its table contains a pre-existing mid-row split whose
  cut edge is now open; bisect attributes the delta to the border commit, with the
  restart-split commit byte-identical across all 89). Everything else reproduces the
  base pixel counts exactly. References untouched.

### DIFF
--- a/packages/docx/src/line-fit-policy.test.ts
+++ b/packages/docx/src/line-fit-policy.test.ts
@@ -1,0 +1,64 @@
+import { describe, expect, it } from 'vitest';
+import { adjustForWidowOrphan, selectLargestFittingEnd } from './line-fit-policy.js';
+
+describe('§17.4.6/§17.3.1.44-.45 selectLargestFittingEnd', () => {
+  it('returns an empty selection when the candidate range is empty', () => {
+    expect(selectLargestFittingEnd(3, 3, 100, () => 1)).toEqual({ end: 3, height: 0 });
+  });
+
+  it('includes an end whose height exactly equals the available height', () => {
+    expect(selectLargestFittingEnd(1, 3, 5, (end) => end === 2 ? 5 : 8)).toEqual({ end: 2, height: 5 });
+  });
+
+  it('returns end === start when the first line does not fit', () => {
+    expect(selectLargestFittingEnd(2, 4, 6, () => 7)).toEqual({ end: 2, height: 0 });
+  });
+
+  it('stops at the first overflow even if a later end would fit', () => {
+    const visited: number[] = [];
+    const heights = new Map([[1, 4], [2, 7], [3, 5]]);
+
+    const selection = selectLargestFittingEnd(0, 3, 5, (end) => {
+      visited.push(end);
+      return heights.get(end) as number;
+    });
+
+    expect(selection).toEqual({ end: 1, height: 4 });
+    expect(visited).toEqual([1, 2]);
+  });
+});
+
+describe('§17.3.1.44-.45 adjustForWidowOrphan', () => {
+  const input = {
+    widowControl: true,
+    start: 0,
+    end: 3,
+    totalLines: 4,
+    belowColumnTop: false,
+  };
+
+  it('drops the last selected line only for a one-line remainder after keeping at least two lines', () => {
+    expect(adjustForWidowOrphan(input)).toEqual({ kind: 'dropLastLine' });
+    expect(adjustForWidowOrphan({ ...input, end: 2 })).toEqual({ kind: 'keep' });
+    expect(adjustForWidowOrphan({ ...input, start: 2 })).toEqual({ kind: 'keep' });
+  });
+
+  it('relocates only a lone first line selected below the column top', () => {
+    const orphan = { ...input, end: 1, belowColumnTop: true };
+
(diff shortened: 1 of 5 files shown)


yukiyokotani:
## Adjudication: vAlign × mid-row split — per-piece (first on-page piece) is correct

Fixture `sample-42` (valign-split-restart): two tables, each a 3-row `vMerge=restart` cell (3 × 300pt atLeast = 900pt span) exceeding the 648pt page-usable height, with a short marker + large slack in the restart cell. Measured from Word's PDF export (`pdftotext -bbox`, 4 pages) plus page renders.

**TABLE 1 (`vAlign=center`):** the merged cell's page-1 piece is just Row 1 (~300pt, y≈174–477pt). The marker `◆ RESTART CELL ◆` lands at **y≈325pt = the centre of that 300pt on-page piece**, and page 2 shows the cell empty. It is NOT at the whole-span centre (span-top+450 ≈ 624pt).

**TABLE 2 (`vAlign=bottom`):** the page-3 piece is Row 1+Row 2 (~600pt). The marker sits at **y≈701pt = the bottom of that on-page piece**, not at the whole-span bottom (≈1000pt, which is on page 4).

Both confirm: Word places the restart-cell content in the **first on-page piece** and applies `vAlign` within **that piece's** height — i.e. **per-piece** alignment. This matches the current implementation. **No change needed.**

PR #934, opened 2026-07-11
fix(docx): suppress space-shrink fit tolerance on justified lines (§17.18.44)

### LINKED ISSUE #698: docx: space-shrink fit tolerance overshoots by one word in narrow justified columns (sample-15 p1)
## Where
`packages/docx/src/renderer.ts` — `layoutLines`' fit decision: `SPACE_SHRINK_RATIO = 0.25` (~L6860). A candidate word whose natural width overflows by less than the line's total inter-word shrink budget (`Σ trailing-space widths × 0.25`) is allowed to fit (Knuth-Plass-style tolerance, absorbing the ~0.1–0.3 px/glyph advance bias between Chromium canvas and Word's layout).

## Symptom
Surfaced by #697 (textbox now uses the main engine, so the tolerance applies to textbox lines too):
- **sample-10 p1 (fix)**: the centered title now stays on one line — matching Word's PDF (it wrongly wrapped to two lines before).
- **sample-15 p1 (regression)**: in the narrow (8.4 cm) *justified* small-print column, the same tolerance pulls "citation" up one line; Word breaks after "…and the full". ~3 wraps shift relative to the Word PDF within the paragraph. VRT within thresholds; references were regenerated with this rendering (user-approved), so the divergence is currently invisible to VRT — this issue is the tracker.

## Analysis direction (no spot heuristics)
The tolerance is one global ratio. In a narrow column the absolute shrink budget is small but so is the line, so a single word ≈ several tolerance units — the ratio's two jobs (absorbing measurement bias vs. emulating Word's justify compression) pull apart at small widths. Candidate approaches to evaluate against BOTH sample-10 and sample-15 Word PDFs (fixture loop from #676 available if a purpose-built bracket is needed):
1. Split the two concerns: a fixed small per-glyph measurement-bias allowance (px-scale, width-independent) + a true justify-shrink budget only for justified paragraphs (Word compresses spaces on justify; ST_Jc semantics).
2. Derive the shrink ceiling from Word-observed behavior with a bracket fixture (vary column width × word length so the fit/no-fit boundary discriminates candidate rules), as done for the 1-inch rule.

## Exit criteria
Both sample-10 p1 (1-line title) and sample-15 p1 (Word's wrap points) match their Word PDFs, without a special case keyed to either sample. Update the regenerated references afterward.

yukiyokotani:
## Problem

`layoutLines` admits a candidate word onto the current line whenever its natural
overflow fits within the Knuth-Plass space-shrink budget
(`Σ trailing-space widths × SPACE_SHRINK_RATIO`, ratio `0.25`). That tolerance is
meant to absorb the Chromium `measureText` vs Word advance-width bias on a line
drawn at (or compressed toward) natural spacing. It was applied to **fully
justified** paragraphs too.

In a narrow justified column the absolute shrink budget is a large fraction of a
single word, so the tolerance pulled one extra word onto a line where Word
breaks earlier, shifting several wrap points within the paragraph relative to the
Word ground truth. (Surfaced after text boxes moved onto the main layout engine,
so the tolerance began applying to text-box lines as well.)

## Fix

ECMA-376 §17.18.44 (`ST_Jc` `both`/`distribute`): Word's line breaker is greedy
and identical for justified and non-justified paragraphs; justification
redistributes the residual slack by **expanding** inter-word spaces and never
compresses a line below its natural width to admit an extra word.

So the space-shrink tolerance is now **suppressed for fully-justified
paragraphs** (`shrinkBudget = 0`) — a justified line breaks at natural fit. The
separate trailing-space-collapse allowance and **all non-justified behaviour are
unchanged**, so a centred/left title that overruns its box only by measurement
bias still stays on one line.

- `line-layout.ts`: thread an `isJustified` flag into `layoutLines`; gate the
  shrink budget.
- `layout-context.ts`: carry `isJustified` on `ParagraphLayoutContext` (classified
  via `jcIsFullyJustified`) so the paginator measure pass and the paint pass make
  the same break decisions.
- `paragraph-measure.ts` / `renderer.ts`: pass the paragraph's justified flag at
  every `layoutLines` call (body paint, autofit measure, and both text-box paths).

## Verification

Measured against the Word PDF ground truth (`pdftotext -bbox`) and the renderer's
`onTextRun`/line geometry at scale 1:

- A narrow (~8.4 cm) justified small-print column now wraps **exactly** as Word
  does — every line boundary matches the PDF, where before one word was pulled up
  per affected line.
- A centred substituted-font title still stays on a **single line** (unchanged).

Empirically, forcing `shrinkBudget = 0` for justified paragraphs reproduced Word's
exact wrapping for the justified column while leaving the centred title on one
row — no special case keyed to any document.

Gates: new focused test `justify-shrink-overshoot.test.ts` (RED before the fix)
plus the updated §17.18.44 draw test; full docx unit suite (1152 tests) and docx
`tsc --noEmit` pass.

Closes #698

### DIFF
--- a/packages/docx/src/justify-shrink-overshoot.test.ts
+++ b/packages/docx/src/justify-shrink-overshoot.test.ts
@@ -8,28 +8,30 @@ import type {
   SectionProps,
 } from './types';
 
-// ECMA-376 §17.18.44 (ST_Jc `both`) — the Knuth-Plass space-shrink fit tolerance
-// (`SPACE_SHRINK_RATIO`) must NOT admit an extra word onto a FULLY-JUSTIFIED line.
+// ECMA-376 §17.18.44 (ST_Jc `both`/`distribute`) — the Knuth-Plass space-shrink
+// fit tolerance (`SPACE_SHRINK_RATIO`) must NOT admit an extra word onto a line
+// that the draw pass will JUSTIFY, and MUST keep doing so on a line the draw pass
+// treats as non-justified. The gate is per LINE, mirroring the paint predicate
+// (renderer.ts: `applyJustify = isJustified && (!endsLogicalLine || stretchLastLine)`):
 //
-// Word's line breaker is greedy and identical for justified and non-justified
-// paragraphs; justification redistributes the residual slack by EXPANDING the
-// inter-word spaces, and (in its default mode) never COMPRESSES a line below its
-// natural width to admit one more word. So on a justified line a candidate word
-// whose natural advance overflows the column must wrap — even if the overflow is
-// smaller than the line's total inter-word shrink budget. The shrink tolerance
-// exists only to absorb the Chromium-`measureText` vs Word advance-width bias on
-// a line that will be drawn at (or compressed toward) its natural spacing, i.e. a
-// NON-justified line (e.g. a substituted-font centred title that Word keeps on one
-// row). See issue #698: in sample-15 p1's narrow (8.4 cm) justified copyright
-// column the tolerance pulled "citation" up onto the "…and the full" line, where
-// Word (PDF ground truth) breaks after "full".
+// - A line that WILL justify (a non-final, non-manual-break line of a
+//   `both`/kashida paragraph, or ANY line of `distribute`/`thaiDistribute`)
+//   redistributes its slack by EXPANDING inter-word spaces — it is never drawn
+//   compressed below natural width, so a candidate word whose natural advance
+//   overflows the column must wrap. Observed Word behaviour (issue #698, PDF
+//   ground truth): in a narrow justified column the tolerance pulled one word up
+//   per affected line where Word breaks at natural fit.
+// - A line the draw pass does NOT justify — the paragraph's true last line and a
+//   line ending at a manual `<w:br/>` (§17.3.3.1) under `both`/kashida, and every
+//   line of a non-justified paragraph — is drawn with the shrink-fit compression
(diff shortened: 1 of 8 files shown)


yukiyokotani:
Addressed the review's High/blocking finding in 66d196f: the shrink suppression is now gated **per line**, mirroring the paint predicate `applyJustify = isJustified && (!endsLogicalLine || stretchLastLine)`.

- The budget only matters for a *marginal* candidate, which necessarily closes its line — so `endsLogicalLine` is decidable at the fit test from the segment queue (exhausted = paragraph's last line; `lineBreak` sentinel next = manual-break line).
- `both`/kashida: true-last and manual-break lines **keep** the old measurement-bias budget (paint draws them non-justified with the promised shrink-fit compression — measure == paint); lines that will justify break at natural fit (observed Word behaviour, issue #698).
- `distribute`/`thaiDistribute`: suppression stays on **all** lines (`jcStretchesLastLine`).
- `stretchLastLine` threaded next to `isJustified` through `ParagraphLayoutContext` and all `layoutLines` call sites; the glued-group pre-flush uses the same per-candidate budget at the group's closing position.
- Acceptance test reworked to pin the full contract, including wrap-point assertions and the true-last/manual-break/distribute cases the review flagged as unverified.

Re-measured against the Word PDF ground truth: the justified narrow-column block still wraps byte-identically on every line, and the centred title case still renders on a single line. Full docx suite (1155) and typecheck pass.

Issue #855, opened 2026-07-05
docx VRT: sample-4 entry declares pageCount 1 but the current renderer paginates 2 pages


yukiyokotani:
Pre-existing discrepancy noticed while adding samples 27-31 to the VRT matrix (PR #854): the committed visual.spec.ts entry for private/sample-4 says pageCount: 1, but the current renderer produces 2 pages — so page 2 is never covered by VRT and has no reference. Determine whether the 2-page pagination matches the Word PDF (if so, bump the entry and generate the page-2 reference; if not, it's a pagination-fidelity bug to fix first).

yukiyokotani:
## Diagnosis: it's a pagination-fidelity bug (Word PDF = 1 page), root-caused to a font-classification defect — not a paginator defect

**Verdict: scenario (b).** `sample-4.pdf` (the Word ground truth) is **1 page** (`pdfinfo` → `Pages: 1`), while the current renderer paginates to **2 pages**. So the committed `visual.spec.ts` entry (`pageCount: 1`) is *correct*; the renderer is wrong. Do **not** bump the entry.

### What overflows
Only the very last block spills to page 2: the "追加情報" (Heading 2) heading + its following paragraph. Page 1 fills to y≈753 pt and the trailing block (18 pt `spaceBefore` from the Heading 2 style + ~21 pt heading + ~19 pt paragraph, bound together by the style's `<w:keepNext/>`) no longer fits under the `fullContentH ≈ 769.9 pt` limit. So this is a **~40 pt cumulative over-estimation**, not a single bad element.

### Where the 40 pt accumulates
Aligning our per-line `y` against the Word PDF's `pdftotext -bbox` positions, the drift is small (~6 pt) through the top table and headings, then jumps inside the **second table** (the "議事項目" agenda table — 5 rows, each `<w:trHeight w:val="1440"/>` with no `hRule` ⇒ `atLeast` 72 pt minimum). Two of its rows measure ~13–16 pt taller than Word's, which is the bulk of the drift.

### Root cause: `<w:family="modern">` → monospace fallback, ignoring `<w:pitch="variable">`
The agenda cells use **Meiryo UI** (`Normal` style `w:eastAsia="Meiryo UI"`). The cell content **width is correct** (our cell ≈ 227 pt vs Word ≈ 226 pt), but our CJK glyphs are measured **~19 % wider**:

[...]

Issue #794, opened 2026-07-05
fonts: generalize metric-compatible font fallback (global fidelity floor)


yukiyokotani:
Global-readiness gap #1 (from the 2026-07-05 review of worldwide positioning).

**Problem**: font substitution quality determines the fidelity floor in every locale, and our current handling is per-font manual work: `font-metrics.ts` carries hand-tuned win-metrics corrections (e.g. Meiryo UI = 1.60em), and documents referencing fonts absent on the client (the known Century case, sample-10) degrade in ways we only discover per-sample. Every language community has its own version of this problem (Calibri/Cambria on non-Windows, PMingLiU/SimSun, Nirmala UI, Sakkal Majalla, …).

**Proposal**:
1. A metric-compatible substitution table for the well-known compatible pairs (Carlito→Calibri, Caladea→Cambria, Liberation Sans/Serif/Mono→Arial/Times/Courier, Noto CJK→MS CJK fonts, etc.), consulted before generic-family fallback.
2. Generalize the metrics layer: instead of per-font constants, measure the actual fallback font's metrics at runtime (fontBoundingBox*) and reconcile against the requested font's declared metrics when available (embedded fonts / os2 from ODTTF path already parsed), so line heights stay Word-like even under substitution.
3. Keep the existing hand-tuned entries as overrides where measurements are known to disagree with Word.

Shared concern across docx/pptx/xlsx → the table and measurement helpers belong in `packages/core` (pure data + functions), consumed by all three renderers.

yukiyokotani:
## Requirement input from #855: the metrics layer needs an *advance-width* axis, not just line height

Root-causing #855 (docx sample-4 paginating 2 pages vs Word's 1) produced concrete measurements that pin down a requirement for the generalized metrics layer proposed here.

**Measured facts** (Chrome, same environment as the VRT):

- Real Meiryo UI (Word's rendering) advances full-width CJK at **~0.84 em** — it is a condensed proportional face (`fontTable.xml`: `family="modern"`, `pitch="variable"`).
- On a system without Meiryo UI, every substitute we can reach — Noto Sans JP, Hiragino Sans, and even the old (incorrect) Courier/monospace fallback — measures full-width CJK at exactly **1.0 em**. All candidate fallback chains are advance-equivalent for CJK.
[...]

PR #942, opened 2026-07-11
fix(docx): gate fontTable modern family to monospace on w:pitch=fixed

yukiyokotani:
## Summary

Refs #855 (partial fix — the issue stays open; see "What this does not fix" below).

`word/fontTable.xml` fonts declared `<w:family w:val="modern"/>` (ECMA-376 §17.8.3.10) were mapped to a `"Courier New", monospace` fallback chain **unconditionally**. Per the spec, `w:family` classifies the typeface *design*, not its spacing — the actual pitch is `<w:pitch w:val="…"/>` (§17.8.3.29; ST_Pitch §17.18.66: `fixed` = Fixed Width, `variable` = Proportional Width, `default` = no information, and an omitted element is assumed `default`).

A proportional "modern" face — e.g. Meiryo UI, which Word's font table records as `family="modern"` + `pitch="variable"` — was therefore measured and drawn against a monospace fallback: fixed-pitch Latin and full 1.0 em CJK advances instead of its proportional sans metrics.

### Changes

- **Parser** (`parser.rs`): `parse_font_table` now also captures `<w:pitch>` into a new `fontFamilyPitches` model field (§17.8.3.29). Only explicitly declared values are recorded; an omitted `<w:pitch>` stays absent (assumed `default` per spec).
- **Classifier** (`line-layout.ts`): the fontTable `modern` branch returns the monospace chain **only when the font's pitch is `fixed`**; `variable` / `default` / absent fall through to the name-pattern / CJK-sans classification, so Meiryo UI now resolves to its proper sans chain (`"Meiryo UI", "Meiryo", "Noto Sans JP", …`). The pitch map rides a WeakMap keyed on the per-document `fontFamilyClasses` identity, so none of the existing `buildFont` call sites change.
- **Renderer** (`renderer.ts`): the four sites that materialize `fontFamilyClasses` register the sibling pitch map via `fontClassesWithPitches`.
- **Tests**: Rust unit test for `<w:pitch>` capture (incl. no synthesis for omitted pitch); three `normalizeFontFamily` tests (modern+variable → not monospace, modern+fixed → monospace, modern+absent → not monospace).

Genuinely fixed-pitch faces keep their monospace fallback: fonts declaring `pitch="fixed"` (e.g. HG Gothic variants in the same document), and Courier/Consolas-style names via the existing name heuristic. pptx/xlsx are unaffected — they classify through the shared name-based core classifier and never consumed `w:family`.

### Measured effect (sample-4, Chrome, same environment as VRT)

| | Meiryo UI resolved chain | CJK advance | page count |
|---|---|---|---|
| before | `…, "Courier New", monospace` | 1.0 em | 2 |
| after | `…, "Meiryo UI", "Meiryo", "Noto Sans JP", …, sans-serif` | 1.0 em | 2 |

The misclassification is fixed (verified via the model's `fontFamilyPitches` and the resolved chain at render time), and Latin/digit runs in `modern`+`variable` faces now render proportionally as Word does.

### What this does not fix

sample-4 still paginates to 2 pages (Word: 1). Word reaches 1 page because real Meiryo UI has a *condensed* CJK advance (~0.84 em); on systems without Meiryo UI installed, the sans substitutes (Noto Sans JP / Hiragino Sans) are em-square (1.0 em — measured identical to the old monospace fallback), so the agenda-table cells still wrap one extra line each and the trailing block still overflows. Closing the remaining gap needs a data-driven advance-width override for metric-compatible fallback (the generalization tracked in #794); `font-metrics.ts` currently corrects line height only.

## Test plan

- `cargo fmt --all --check` / `cargo clippy --all-targets -- -D warnings` / `cargo test` (317) — pass
- WASM rebuilt via the package script (`fontFamilyPitches` verified present in the built artifact)
- `npx vitest run packages/docx` — 152 files, 1175 tests pass
- `pnpm --filter @silurus/ooxml-docx typecheck` — pass
- docx VRT triage — results posted as a PR comment

### DIFF
--- a/packages/docx/parser/src/parser.rs
+++ b/packages/docx/parser/src/parser.rs
@@ -346,7 +346,7 @@ pub fn parse(zip: &mut Zip) -> Result<Document, String> {
         })
         .unwrap_or_else(|| "word/fontTable.xml".to_string());
     let font_table_xml = read_zip_string(zip, &font_table_path).unwrap_or_default();
-    let font_family_classes = parse_font_table(&font_table_xml);
+    let (font_family_classes, font_family_pitches) = parse_font_table(&font_table_xml);
     // ECMA-376 §17.8.3.3-.6 — embedded fonts. The `<w:embed*>` r:ids resolve
     // through the fontTable part's OWN relationships (`word/_rels/fontTable.xml.rels`
     // when the part is `word/fontTable.xml`): insert `_rels/` before the filename
@@ -412,6 +412,7 @@ pub fn parse(zip: &mut Zip) -> Result<Document, String> {
         major_font,
         minor_font,
         font_family_classes,
+        font_family_pitches,
         embedded_fonts,
         revisions,
         comments,
@@ -1017,7 +1018,7 @@ fn find_rel_target(rels_xml: &str, type_suffix: &str) -> Option<String> {
     None
 }
 
-/// Parse `word/fontTable.xml` and build a map from font name to family class.
+/// Parse `word/fontTable.xml` and build maps from font name to family class and pitch.
 ///
 /// ECMA-376 §17.8.3.10 defines `<w:font w:name="…"><w:family w:val="…"/></w:font>`
 /// where val is one of: `roman` (serif), `swiss` (sans-serif), `modern`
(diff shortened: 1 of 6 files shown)


yukiyokotani:
## VRT triage (docx suite, branch vs base = main @ 8312297)

Ran the full docx VRT from this branch's worktree and an identical base run from a main checkout (same machine, same references, port isolation verified). Per-page match% compared across all 80 rendered pages (sample-10 skipped — no local reference).

### Branch-originated diffs: exactly 3 pages

| page | base match | branch match | delta | modern+variable fonts involved |
|---|---|---|---|---|
| private/sample-4 p1 | 99.9% | 94.5% | -5.4 | Meiryo UI |
| private/sample-2 p1 | 98.9% | 97.2% | -1.7 | MS PGothic, Meiryo UI |
| private/sample-30 p3 | 81.9% | 84.6% | +2.7 | Malgun Gothic, Dotum |

**All other 77 pages are pixel-identical to base.** The suite's broader failures (sample-5/26/28/29/31, demo/sample-1) reproduce identically on main — pre-existing reference drift, not from this branch. The three moved pages are exactly the samples whose font tables declare `family="modern"` + `pitch="variable"` faces — matching the predicted blast radius; every other sample's "modern" fonts are `pitch="fixed"` (Courier New, Consolas, HG Gothic, MS Gothic, …) and keep their monospace chain byte-identically.

### Direction (vs the Word PDF ground truth)

The match% drops for sample-2/-4 are against the *pre-fix renderer's own output* (the reference PNGs), which any behavior change necessarily diffs against. Against Word:

- **sample-4 p1**: the base rendering shows fixed-pitch Courier gaps in mixed runs (e.g. `開始時刻 – 終了時刻` with wide dash/space advances); the branch matches Word's tight proportional spacing, and placeholder lines fit more text per line, as in the PDF. Toward Word.
- **sample-2 p1**: `請求書番号[100]` / `日付: [日付]` — base renders brackets/digits/colons at Courier fixed pitch (wide gaps); branch matches Word's tight spacing. Toward Word. (MS PGothic's "P" literally means proportional — the monospace mapping was flatly wrong for it.)
- **sample-30 p3**: match improved (+2.7) and region-level distance to the PDF page decreased. Toward Word.

Region-level mean-luminance distance to the PDF is noisy for s2/s4 (the CJK face is a substitute, not real Meiryo UI, in both base and branch, so shifted glyph positions dominate the metric); the structural spacing evidence above is the interpretable signal, and it is consistently toward Word.

### Follow-up

After merge, the local references for private/sample-2 p1, sample-4 p1, sample-30 p3 need regeneration (`UPDATE_REFS=1`) to adopt the corrected classification — flagged for the maintainer per the reference-update policy.

Issue #796, opened 2026-07-05
qa: multilingual verification corpus (Arabic / Thai / Korean / Cyrillic) with Office-PDF ground truth


yukiyokotani:
Global-readiness gap #3 — the structural one. Our verification methodology (render → pixel-compare against Word/Excel/PowerPoint PDF output) is language-neutral, but the private corpus is Japanese-heavy. Every new sample keeps surfacing real latent bugs (sample-30 scatter, sample-12 caption, sample-26 vertical text), so coverage breadth directly bounds claimed fidelity.

**Plan**:
1. Author small document sets (docx + xlsx + pptx, a few documents each) in **Arabic** (RTL + kashida justification once #724 lands), **Thai** (dictionary line breaking), **Korean** (hangul word wrap), **Cyrillic** (baseline sanity for non-Latin LTR) — exercising paragraphs, tables, lists, headers/footers, number formats, charts with localized labels.
2. Export ground-truth PDFs from real Office (needs the maintainer's Office environment; possibly language packs for locale-default fonts/styles — coordinate with the maintainer for this step).
3. Store alongside existing `private/sample-*` conventions (gitignored, VRT local-only), wire into the existing VRT + pdftotext-bbox pipelines unchanged.
4. Triage whatever falls out into spec-first fixes (expected: line-breaking, justification, field/number localization, font-substitution gaps → cross-link #724 and the metric-compatible-fallback issue).

The RTL foundation (UAX#9 bidi, cs-axis mirroring) already exists, so the Arabic corpus likely has the highest yield-per-effort.

yukiyokotani:
Corpus phase complete: 8 samples authored (docx 36-39 Arabic/Thai/Korean/Cyrillic, pptx 17-18, xlsx 31-32, all local per the private-corpus convention), maintainer-exported Word/PowerPoint PDFs collated (xlsx stays on the renderer-self-baseline convention), and the full pixel/structure triage ran. Outcome: the layout engine is script-neutral (Cyrillic matched Word within ~1px on all lines); genuine defects were split out as #928 (resolved: not reproducible, characterization tests in #936), #929, #930, #931 (fixed), plus adjudication follow-up #937. Font-substitution effects route to #794 and the Thai dictionary-breaking question to #797. Closing — the corpus and pipeline are now standing infrastructure; follow-on fixes are tracked in the issues above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #943, opened 2026-07-11
feat(docx): true kashida justification via U+0640 tatweel insertion (§17.18.44)

### LINKED ISSUE #724: docx: true kashida justification (U+0640 tatweel insertion) for jc=lowKashida/mediumKashida/highKashida
Follow-up from the WD8 fix (fix/docx-dropped-runs): `jc` values `lowKashida` / `mediumKashida` / `highKashida` (§17.18.44 ST_Jc) are currently **mapped to the ordinary justify path** — line slack is distributed as inter-word spacing, which is the documented approximation.

True kashida justification in Word elongates Arabic words by inserting/stretching U+0640 (tatweel) at valid connection points instead of widening spaces, with the three values controlling elongation aggressiveness. Implementing it needs:

- Arabic joining-type analysis to find valid tatweel insertion points (after connecting letters, not at word edges or non-joining pairs)
- distribution of line slack across insertion points (per jc level)
- measurement/paint parity for the inserted tatweel glyphs (measure == paint invariant)
- interplay with the existing UAX#9 bidi pipeline and CJK justify kernel

The mapping site is `packages/docx/src/bidi-line.ts` (`resolveAlignEdge`, `jcIsFullyJustified`), marked with a comment referencing this issue.

yukiyokotani:
Closes #724

## What

Implements TRUE kashida justification for `w:jc` = `lowKashida` / `mediumKashida` / `highKashida` (ECMA-376 §17.18.44 ST_Jc): the line slack of an Arabic justified line is filled by **elongating words with inserted U+0640 tatweel at valid cursive joining points**, instead of widening inter-word spaces (the previous documented approximation).

## Why the parser change too

`normalize_align` in the Rust parser collapsed all four §17.18.44 variant values (`lowKashida` / `mediumKashida` / `highKashida` / `thaiDistribute`) into the `left` fallback, so a **real file** carrying them rendered left-aligned — the renderer's existing mapping (`bidi-line.ts`, PR for WD8) was only reachable from direct-model unit tests. The first commit passes them through verbatim (pinned by Rust unit tests); this also activates the already-implemented `thaiDistribute` justification for real Thai documents.

## Design

- **`arabic-joining.ts` + generated Joining_Type table** (UCD 17.0.0 `extracted/DerivedJoiningType.txt`, checked-in source + `scripts/gen-arabic-joining.mjs`, mirroring the core vertical-orientation generator): a tatweel may be inserted between letters A|B when A joins to the following (jt ∈ D/L/C) and B joins to the preceding (jt ∈ D/R/C), skipping Transparent harakat for adjacency and excluding the mandatory lam-alef ligature.
- **`kashida-justify.ts`** — pure, unit-tested allocator. Round-based whole-line round-robin: every eligible point receives its first tatweel before any receives a second. The per-opportunity cap encodes the three aggressiveness levels (low = 1, medium = 2, high = unbounded). Each insertion is accepted only when the re-measured augmented string grows by a positive delta that still fits the remaining slack, so the fixed tatweel advance never overshoots (measure == paint by construction; contextual-shaping-safe because the WHOLE augmented string is re-measured, never `count × width(tatweel)`).
- **`renderer.ts`** — body and Word-textbox justify paths share one allocation helper. The measurement callback uses the delta form pinned to layout's `measuredWidth` and grows it under the segment's own font / kerning / `w:w` scale / char-spacing model. Kashida slack is applied first; only the **residual** feeds the existing space/CJK distributor. An augmented Arabic word is painted as **one `fillText`** so the browser's contextual shaping keeps the tatweels connected (never sliced through `justifiedPiecePositions`); the advance delta rides the existing `internalStretch`, so decorations (highlight / shading / border / underline / strike, ruby centring) and the pen advance stay exact. Semantic consumers (`onTextRun`, find/selection, emphasis marks) keep the ORIGINAL run text.
- **Fallbacks**: a kashida-jc line with no eligible joining point (Latin/CJK content) falls back to the ordinary `both` distribution with the full slack; negative slack uses the existing spaces-only compression (kashida never compresses); last-line policy matches `both` (natural).

## Observed-range honesty / Word adjudication

Word's exact kashida algorithm is undocumented. The generic Unicode joining rule + lam-alef exclusion establishes where a join CAN survive, not Word's glyph-level priority order (Uniscribe's `SCRIPT_JUSTIFY` priorities, e.g. Seen-class first, are not observable from Canvas). The low/medium/high = 1/2/∞ cap is a provisional monotonic policy matching the spec's "slightly longer / medium / widest possible" wording — flagged in code comments as pending adjudication.

**`packages/docx/scripts/gen-kashida-fixture.mjs`** (dependency-free, deterministic) generates `packages/docx/tests/fixtures/kashida-justification.docx` — an RTL Arabic paragraph under `both` + the three kashida levels (`*.docx` binaries stay gitignored by repo policy). Rendering that file in Word and exporting a PDF next to it gives the ground truth to adjudicate insertion sites and level aggressiveness in a follow-up.

## Verification

- New tests: 23 (joining analysis 8, allocator 7, renderer end-to-end 8 incl. `both`/`left` negative controls).
- Full docx suite: **1200/1200**; root `pnpm typecheck` clean; `cargo fmt --check` / `clippy -D warnings` / `cargo test` (327) clean.
- WASM round-trip probe: parsed alignments of a kashida fixture now surface as `lowKashida`/`mediumKashida`/`highKashida` (previously all `left`).
- Private corpus before/after paint-diff probe (same renderer, old-vs-new alignment mapping):
  - sample-28 (Arabic; carries `lowKashida` only in **unused** styles): 0 pages differ — byte-identical.
  - sample-36 (Arabic corpus, `both`/`right` only): 0 pages differ — byte-identical; the #796 D3 drift is untouched (no kashida jc there to improve).
  - sample-29 (Thai; 15 `thaiDistribute` paragraphs): pages now justify. Word PDF ground truth confirms the direction — e.g. p.2 has 15 lines sharing an identical justified right edge (xMax 532.2pt) where we previously rendered ragged-left.
- Non-kashida output is byte-identical by construction: every new code path is gated on the three kashida alignment strings.


yukiyokotani:
## Adjudication: kashida positions & elongation (fixture sample-40)

`sample-40` (kashida-priorities): a 2.5in column (large slack) with the same Seen/Sheen-rich Arabic sentence set as `both` (control) / `lowKashida` / `mediumKashida` / `highKashida`. Word inserts **literal U+0640 tatweel** glyphs, extracted with `pdftotext -raw` and analysed by reconstructing each tatweel run's join (prev→next glyph).

### (a) Insertion positions — specific-glyph PRIORITY, stable across levels (NOT uniform round-robin)
Kashida lands at **one priority join per word** (25 runs across the paragraph — roughly one per word), and the **join set is identical** for low/medium/high — only the elongation grows. It is emphatically not a uniform round-robin over every eligible join (each word has several joins; only one is elongated). The joins Word chose (constant across all three levels):

```
ﺳ→ﻠ  ﺳ→ل  ﻓ→ﻲ  ﺷ→ﻔ  ﺳ→ﺔ  ﺳ→ﺗ  ﺑ→ﺔ  ﺻ→ﻠ  ﺑ→ﯾ  ﺣ→ر
```
(after seen/sheen/sad and before final letters — the classic Uniscribe kashida priority).

### (b) Elongation — monotonic, low is a fixed minimum
Total tatweel glyphs: **both=0, low=50, medium=106, high=131.** Per-kashida run length:
- low: uniformly **2** tatweel (histogram `{2:25}`) — a fixed minimal elongation; wrapping stays identical to the `both` control (6 lines).
- medium: `{2:5, 4:4, 5:16}` (up to 5), 7 lines.
- high: `{3:5, 4:8, 5:12, 8:3}` (min 3, up to 8), 7 lines.

### Relation to the current implementation
Current = uniform round-robin over **all** joins with a per-opportunity cap 1/2/∞. It already agrees with Word on "positions are stable across levels", but differs on **which** joins (all joins vs ~one priority join per word) and on the elongation model.

**Recommendation — defer the position change.** ECMA-376 §17.18.44 defines only the `ST_Jc` values; it does not specify the kashida priority ranking or elongation amounts (those live in Uniscribe / a Unicode Arabic-justification note, and are font-dependent). Reverse-engineering the full priority table from this single sentence would be a sample-fitted heuristic, which this repo's spec-first / no-heuristic rules forbid. The adjudication above is definitive on the *behaviour*; adopting Word's priority faithfully should be a deliberate design decision backed by more fixtures rather than a fit to sample-40. Not implemented in PR #952.

PR #949, opened 2026-07-11
fix(docx): place RTL trailing whitespace on the leading side for skia parity

### LINKED ISSUE #929: docx: Arabic/RTL inter-word space collapses at run boundaries
Found via the #796 multilingual corpus triage (private fixture; no content quoted).

When two words of an RTL paragraph fall in adjacent runs, the visual inter-word space disappears — the words render touching — whereas Word shows a normal space. The trailing space's advance is measured into the run width, but the neighbouring run is positioned flush at the run edge, so the gap vanishes. Reproduced on several two-word labels and table cells (gap ≈ 0px vs a full space width). RTL glyph order itself is correct (UAX#9 reorder intact).

Font-independent (reproduces with the substituted face). Possibly interacts with the kashida/justification work tracked in #724 but is a distinct positioning defect.

yukiyokotani:
## Problem

An inter-word space at a segment boundary in an RTL (`<w:bidi>`, §17.3.1.6) paragraph disappeared, rendering the two words touching, whereas Word shows a normal space. Most visible on two-word labels and table cells (e.g. an Arabic name in a cell), whose single inter-word gap collapsed.

## Root cause

The paint pass drew each text segment's whole string — including its trailing inter-word space — with one `fillText`, relying on `ctx.direction='rtl'` to reorder that trailing space onto the segment's physical **left** (toward the next reading word). That reordering is **backend-dependent**: Chrome's Canvas performs it, but skia-canvas — the Node/server rendering backend used for VRT and the MCP server — left-anchors the logical string and leaves the trailing space on the physical **right**. In an RTL paragraph the reading-next word sits to the physical left, so the space landed on the wrong (outer) side and the word rendered flush against its neighbour; the gap vanished.

This is why the browser viewer looked correct while the server-rendered output collapsed the gap.

## Fix

Position the edge whitespace explicitly instead of relying on the backend's per-string bidi. For an RTL-direction segment ending in whitespace, draw the trailing-whitespace-trimmed glyphs anchored at `boxLeft + trailingWhitespaceAdvance` so the whitespace always occupies the box's left — identical output in Chrome and skia. Mirrors the existing fitText region-end RTL pad shift (`fitDrawX`).

LTR segments and the non-bidi fast path keep the same `fillText` string and anchor (byte-identical). Run decorations, `onTextRun`, and the pen advance stay on the untrimmed box (`x` / `spanW`).

## Verification

- **Unit (committed, backend-independent):** `packages/docx/src/rtl-run-boundary-space.test.ts` — an RTL trailing-space word draws its trimmed glyphs shifted one space-width to the right (space on the box left); LTR control unchanged. Fails against the pre-fix renderer.
- **skia backend (empirical):** rendering the affected two-word cells through skia-canvas shows the words touching before the fix and a full inter-word space restored after (matching Word).
- **Regression:** full docx unit suite (1203 tests) unchanged; the change is gated to RTL trailing-whitespace segments, so all other content is byte-identical (the browser VRT is unaffected — Chrome already reordered correctly).

Closes #929

### DIFF
--- a/packages/docx/src/renderer.ts
+++ b/packages/docx/src/renderer.ts
@@ -8126,6 +8126,44 @@ function drawParagraphLine(li: number, c: ParagraphLineDrawCtx): void {
         //   4. Neither: a single fillText (the common path).
         const segmentGridDeltaPx = segmentCharacterGridDeltaPx(s, drawGridDeltaPx);
         const segGridDelta = gridSegDeltaPx(drawText, segmentGridDeltaPx);
+        // ECMA-376 §17.3.1.6 `<w:bidi>` (issue #929) — a segment's TRAILING
+        // whitespace (an inter-word space at its logical end) must sit on the
+        // segment's physical LEFT under an RTL visual frame, toward the next
+        // reading word. Canvas is asked to do this via `ctx.direction='rtl'`, but
+        // that is BACKEND-DEPENDENT: Chrome reorders the trailing space to the
+        // left, whereas skia-canvas (the server/VRT/MCP rendering backend)
+        // left-anchors the logical string and leaves the space on the physical
+        // RIGHT — so the space lands on the wrong (outer) side and the word renders
+        // FLUSH against its reading-next neighbour (the gap collapses; most visible
+        // as a two-word label / table cell where the single inter-word gap is lost).
+        // Position the whitespace EXPLICITLY instead: draw the trailing-whitespace-
+        // TRIMMED glyphs shifted rightward by the whitespace advance so the space
+        // occupies the box's LEFT — identical output in both backends. This mirrors
+        // the fitText region-end RTL pad shift above (`fitDrawX`). LTR segments and
+        // the non-bidi fast path keep `glyphText===drawText` / `glyphDrawX===x`
+        // (byte-identical). Decorations, `onTextRun`, and the pen advance stay on
+        // the untrimmed box (`x` / `spanW`).
+        let glyphText = drawText;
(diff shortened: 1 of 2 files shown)


yukiyokotani:
Review follow-up pushed (832007b), addressing both findings:

1. **[high] fitText branch bypass** — the §17.3.2.14 branch now draws the trimmed `glyphText` at `fitDrawX + rtlWsShiftPx`, reserving both the region-end pad and the whitespace advance (glyph width + per-gap share) on the left. Added a fitText×RTL trailing-space case (fails against the previous commit). The docGrid / justified split-piece branches are documented as exempt: they require CJK inside the segment, which resolves to an even (LTR) level, so an RTL-direction segment cannot reach them outside the rtl-marked EA-punctuation corner where bidi justification is already approximate.
2. **[medium] w:w + w:spacing measurement** — the shift is now derived from the same `segAdvanceWidth` authority the measure pass used for the box (natural × w:w scale + per-cp pitch), instead of re-measuring under a paint letterSpacing (which scaled the fixed pitch). The test mock now honors letterSpacing per code point and tracks translate/scale, so the pitch-scaling mistake is detectable; the new w:w+spacing case asserts shift = authority Δ (8, not the inflated 11) and the right-edge identity.

Verification: both new cases Red against a158489 / Green now; full docx suite 1205 green; skia re-render confirms the affected table cells keep a full inter-word space and a synthetic RTL fitText pair places the widened slot (per-gap + space share) exactly between the words in reading order (pixel-measured), leading glyph on the region's right edge.

Issue #804, opened 2026-07-05
docx: determine Word's behavior for a continuous-section w:start restart observed at a spillover page boundary


yukiyokotani:
Residual from the PR #800 review (WD2 page numbering). When a `<w:type w:val="continuous"/>` section carries an explicit `<w:pgNumType w:start="N"/>` and its content spills across a page boundary, our `computePageNumbering` resets the counter to N on the first page whose top that section owns. Whether Word does the same (vs. ignoring the restart for continuous sections entirely, vs. applying it retroactively from the section's first page) is undetermined: the only real fixture (sample-13, continuous start=2) has a start value that coincides with natural continuation (1+1=2), so all candidate semantics produce identical output. A synthetic probe (continuous start=50 spilling) yields [1, 50, 51] under the current implementation.

To settle: author a docx with a continuous section whose start differs from natural continuation (e.g. start=50) and enough content to spill, export the PDF from real Word (maintainer environment), and compare footer numbers. Then either keep the current semantics or adjust `computePageNumbering` — and replace the code-comment residual note with the spec/ground-truth citation.

Related: multilingual/edge-case corpus effort #796 (same "needs maintainer Office environment" dependency).

PR #1050, opened 2026-07-17
refactor(docx): make page flow an immutable state machine

yukiyokotani:
## Summary

- replace renderer-owned body pagination state with immutable page/column transitions and retained `DocumentLayout` ownership
- keep main-thread and render-worker layout variants keyed by the same normalized inputs and service fingerprints
- retain section regions, page geometry, numbering, bookmarks, tables, drawings, and page-local ownership through one canonical producer
- remove obsolete pagination stamps, the alternate retained table painter, and the dead section-break spacer helper
- harden the DOCX layout boundary baseline against duplicate JSON keys

## Specification and compatibility

- ECMA-376 Part 1 §17.18.77: `nextColumn` uses the exact following same-page column when the incoming section can safely own that physical band
- when the active region has no following column, including a single-column section, continue in the incoming section on the following page; this preserves the established pre-cutover non-continuous section behavior instead of rejecting the document
- incompatible same-page page extents, writing modes, block bands, grids, or physical column ownership remain fail-closed with clone-safe structured diagnostics
- ECMA-376 §17.6.11 and §17.6.13 do not define a US Letter/one-inch default for omitted page geometry; non-continuous omissions preserve the established document page box, while continuous sections inherit omitted page-level facts from the following resolved section
- compatibility-only behavior is registered with concrete regression-test references rather than fabricated Office version/platform metadata

## Review fixes

- restore production bookmark propagation from parser paragraphs through retained paragraph slicing and page metadata
- align eager default-layout failure timing between main and worker modes and preserve structured error fields across the worker bridge
- retain only the load-time default plus one reusable non-default layout variant
- add true-flag coverage for mirror margins, top/RTL gutter behavior, and imposition modes
- restore real-model continuous page-number restart and rendered `PAGE` field coverage
- remove dead retained table paint code and obsolete boundary-baseline declarations
- move continuous-section mark-spacing compatibility behind the registered compatibility projection
- avoid full-suite timeout contention in the Node DOCX find/highlight integration test without weakening its real-render assertions

## Verification

- `pnpm test` — 5,469 passed, 26 skipped
- `pnpm typecheck`
- `pnpm lint`
- `pnpm lint:test`
- `cargo test -p docx-parser` — 415 passed
- `node --test scripts/check-docx-layout-boundaries.test.mjs`
- `node scripts/check-docx-layout-boundaries.mjs`
- `node scripts/check-docx-public-api.mjs`
- `pnpm build`
- `pnpm smoke` — 18 passed across Chrome, Firefox, and WebKit


yukiyokotani:
## Independent architecture review — retained body layout cutover

**Verdict: REQUEST CHANGES.** The cutover architecture itself is sound and well-gated (single canonical producer, deep-frozen retained layout, explicit coordinate-space tags, paint purity enforced statically, no silent fallback found). But CI is red on this PR head (the merge-base commit on main is green), and two functional regressions were verified in the diff.

### Blockers — CI is red at `e1e65e2`

1. **Smoke red in all 3 browsers.** `tests/smoke/layouts.spec.ts:117/127/136` (DOCX ScrollView / ThumbnailGrid / MasterDetail): webkit/firefox render 2+ blank pages (`countInkedCanvases < count-1`); chrome times out waiting for load. The merge-base run is green, so this is an end-to-end rendering regression of the cutover and must be root-caused before merge.
2. **typecheck red — the PR violates its own boundary rule.** `packages/docx/src/paint/canvas-page.ts:8` value-imports `orderedPagePaintEntries` from `../layout/page-graph.js`; the PR's own `no-docx-layout-in-paint` ast-grep rule reports it at error level. The function is a retained-data projection (it resolves the precomputed `layers.paintOrder`), so either relocate it to a paint-consumable retained-access module or explicitly allow it in the rule — either way CI must be green.
3. **rust red.** The committed `packages/docx/parser/src/parser.rs` is not `cargo fmt`-clean (repo gate: fmt + clippy).

### Major

4. **Bookmark navigation is silently broken end-to-end (main + worker).** `packages/docx/src/layout/page-factory.ts:384-390` derives `page.bookmarkStarts` from `placement.bookmark`, but no production code ever sets that field (`layout/paragraph.ts` never emits it), and nothing reads the parser-emitted `DocParagraph.bookmarks` anymore (exhaustive grep: zero non-test references). Result: `DocxDocument.getBookmarkPage()` always returns `undefined`, `<w:hyperlink w:anchor>` clicks are inert, and the worker meta `bookmarkPages` is always empty (ECMA-376 §17.13.6.2 / §17.16.23). Tests miss it because they stub `bookmarkStarts` directly. This behaviorally breaks a public API despite unchanged declarations. Fix: stamp paragraph bookmarks into the retained graph during acquisition (including inkless paragraphs) and add one real-model test asserting non-empty `bookmarkStarts` through `layoutDocument`.
5. **A `nextColumn` section break can now hard-fail the whole document.** The parser now passes `nextColumn` through (`parser.rs` `section_break_element`; the base coerced it to `nextPage`), and `packages/docx/src/layout/paginator.ts:259-289` throws `UnsupportedPageFlowTransitionError` whenever no following column exists — which is always the case in single-column sections. No production caller catches it: in worker mode `parse` rejects (the document won't open at all); in main mode `pageCount`/`renderPage` throw. §17.18.77 leaves the missing-column case undefined, but the codebase's established posture (parse error → explicit error page) shows whole-document rejection is not the intended degradation. Fix: fall back to a next-page transition (matching base behavior) or catch at the `layoutDocument` boundary into a diagnostic.
6. **Undisclosed behavior change: geometry-less non-continuous sectPr now gets US Letter defaults.** `packages/docx/src/parser-model.ts:1254-1262` (+ the rewritten expectation in `per-section-page-geometry.test.ts`: body-geometry fallback → 612×792). Two issues: (a) the continuous-only inheritance is backed by §17.18.77 (ST_SectionMark), not the cited §17.6.22; (b) ECMA-376 defines no schema defaults for `pgSz`/`pgMar`, so 612×792 is an application default, not "ECMA defaults" as the comment claims. Practical failure: an A4 document containing a geometry-less `nextPage` sectPr now renders a Letter-sized page where the previous engine rendered A4. This directional change is not mentioned in the PR description, and the non-continuous branch has no test. Fix: correct the citations, record the default as application-defined (or restore body-geometry inheritance after a Word ground-truth check), state the change in the PR body, and add a test.
7. **New mirror/gutter/book-fold geometry logic is uncited, untested, and parity-inconsistent.** `packages/docx/src/layout/context.ts:107-126` introduces three undocumented decisions (gutter suppressed under any imposition flag; `gutterAtTop && !mirrorMargins`; mirror swap keyed to physical `pageIndex % 2 === 1`) while even/odd header selection keys off `displayPageNumber % 2` (`layout/header-footer-reserve.ts:31-33`). With `<w:pgNumType w:start>` shifting parity, margin mirroring and even/odd headers disagree on the same page; Word keys both off the displayed number. Relevant clauses: §17.6.12, §17.15.1.57 (mirrorMargins), §17.15.1.44 (gutterAtTop), §17.15.1.79 (rtlGutter). No test sets any of these flags true. Fix: key mirroring to the display number, cite the basis per branch, and add true-flag tests.
8. **End-to-end continuous-restart page-numbering pins were deleted without full replacement.** The deleted `page-numbering.test.ts` pinned the shared-page restart sequence and the "restart island that owns no page top" case (#804, §17.6.12) through real measurement to the rendered PAGE field. The replacements in `layout/body-page-numbering-production.test.ts:296,432` use hand-built inputs with stub heights, and the island case has no replacement at all; the dead helper `continuousSpilloverDoc` remains in `page-number-field-render.test.ts:151`. Fix: reinstate one real-model spillover PAGE-field test (the dead helper is ready-made) and re-pin the island case.

### Minor

9. **Dead alternate table-paint path left in `renderer.ts`.** `:822-826`, `:1972-2076`: `retainedNestedTableResolver`/`retainedTablePainter` are constructed on every render but never invoked (their only call site was deleted by this PR). ~105 lines duplicating `paint/canvas-table.ts` responsibilities, including a second vertical-rotation transform — a dormant drift hazard. Delete.
10. **Unbounded layout-variant cache keyed by raw `currentDateMs`.** `layout/variant-store.ts:32,60-64` + `layout/options.ts:28-35`: a caller passing `currentDate: new Date()` per render re-paginates the document and retains every `DocumentLayout` forever. Normalize the key when the layout has no date-sensitive fields (the session already tracks `hasPaginationFields`) or bound/evict the map.
11. **The worker "parity" test never executes `render-worker.ts`.** `render-worker-layout-parity.test.ts:64-159` stubs the bridge and builds both "sides" inside the test. Actual parity is guaranteed by the static `WORKER_LAYOUT_SELECTION` check, not by this test; rename it or add a run through the real worker module with a non-trivial document.
12. **Boundary-checker hygiene.** `scripts/check-docx-layout-boundaries.mjs:489`: `assertBodyPaintConsumesRetainedLayout` is now vacuous (`renderBodyElements` no longer exists) and ~900 lines of computePages-era normalization machinery are dead at HEAD; `scripts/docx-layout-boundary-baseline.json:549-550` contains a duplicate key that will churn on the next regeneration. Follow-up cleanup keeps the checker auditable.
13. **Compatibility-evidence records don't resolve to fixtures.** The `syntheticFixtureId`s in `layout/body-pagination-compatibility.ts` match no file or test name, and the section-mark spacing interop roles (`layout/body-layout-input.ts:26-31,202-232`) bypass `defineCompatibilityRule` entirely — their recorded evidence is stranded on the now-dead `isSectionBreakSpacerAt` (`renderer.ts:4890`). Register the roles, link real fixture/test IDs, and delete the dead helper.
14. **Worker vs main failure-surface divergence.** The worker builds `defaultLayout` eagerly at `parse` (a failing document fails at load), while main fails lazily at first layout use; page-index validation moved from a sync throw to an async bridge rejection (`document.ts:498-533`). Intentional-looking, but worth documenting or aligning.

### Verified positives

- Single canonical producer (`layout/body-paginator.ts#paginateBody`), reached identically from renderer, worker, and `document.ts`; legacy `computePages`/`paginateDocument` fully deleted; reintroduction blocked by static checks whose baseline strictly shrank.
- Retained layout is deep-frozen and invariant-asserted; no post-production mutation found; coordinate spaces are tagged end-to-end and the region matrix is applied exactly once (pinned by `vertical-coordinate-cutover.test.ts`).
- No migration flag or silent fallback; convergence loops throw on fingerprint cycles instead of magic iteration caps; no sample-specific conditionals in the new modules.
- Public API declarations unchanged (`index.ts` diff is comment-only; the deleted `PaginatedBodyElement` was never exported); Rust↔TS wire additions are all optional-with-defaults, so old-WASM/new-TS pairing degrades gracefully (except item 5).
- The large legacy-test rewrites did not weaken assertions: expectation changes are coordinate-space corrections, with item 6 as the single directional exception.

Items 1–3 (red CI) plus 4–5 (verified regressions) need another pass before merge; the retained-engine architecture itself is in good shape.

PR #1055, opened 2026-07-19
refactor(docx): preserve complete text box content

yukiyokotani:
## Summary

- preserve the complete ordered `w:txbxContent` block stream for both WPS and VML text boxes
- reuse the existing paragraph and table parser payloads, including nested tables, images, charts, hyperlinks, and part-local relationships
- retain unsupported schema-permitted block children as ordered diagnostic markers
- keep the stable public `ShapeRun` declaration and existing `textBlocks` rendering behavior unchanged
- snapshot the parser-only wire at the TypeScript parser-model boundary

## Why

ECMA-376 defines `CT_TxbxContent` as `w:EG_BlockLevelElts+`, but the existing parser reduced text boxes to a paragraph-only compatibility projection. That discarded tables and other authored block children before the retained layout engine could consume them.

This is the acquisition-only checkpoint for issue #1037 Series B. A follow-up PR will migrate text-box layout to the shared block layout path; this PR deliberately does not change rendering behavior yet.

The complete wire uses an isolated numbering snapshot, while the live compatibility state advances only through paragraphs visited by the pre-B2 `textBlocks` path. This keeps the wire internally coherent without allowing newly acquired table content or suppressed VML textPath bodies to change current rendering or following list numbers. WPS and VML share the same block parser and recursion-depth behavior.

## Verification

- `cargo test -p docx-parser` (424 passed)
- `cargo clippy -p docx-parser --all-targets -- -D warnings`
- `pnpm build:wasm`
- `pnpm typecheck`
- `pnpm test` (5,489 passed, 26 skipped)
- `node scripts/check-docx-public-api.mjs`
- `node scripts/check-docx-layout-boundaries.mjs`
- `git diff --check`

Part of #1037.

### DIFF
--- a/packages/docx/parser/src/numbering.rs
+++ b/packages/docx/parser/src/numbering.rs
@@ -124,7 +124,7 @@ enum CounterKey {
     OrphanNum(u32),
 }
 
-#[derive(Default)]
+#[derive(Default, Clone)]
 pub struct NumberingMap {
     /// abstractNumId → [level0..level8]
     abstract_nums: HashMap<u32, Vec<LevelDef>>,
--- a/packages/docx/parser/src/parser.rs
+++ b/packages/docx/parser/src/parser.rs
@@ -2862,7 +2862,7 @@ fn parse_paragraph(
     table_style_id: Option<&str>,
     field: &mut FieldState,
 ) -> DocParagraph {
-    parse_paragraph_cond(
+    parse_paragraph_cond_at_depth(
         node,
         style_map,
         num_map,
@@ -2873,6 +2873,7 @@ fn parse_paragraph(
         table_style_id,
         None,
         field,
+        DepthGuard::root(),
     )
 }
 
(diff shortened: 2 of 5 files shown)


yukiyokotani:
Final review checkpoint for `2a43a3af`:

- Claude Fable 5 ultrareview completed against the public final branch diff (`origin/main..HEAD`, 5 files) with **no findings**.
- The acquisition-only numbering compatibility issue found during the first review run was fixed before this final review: complete wire numbering is isolated from the still-active legacy projection, including newly acquired table blocks and suppressed VML textPath bodies.
- Fresh local verification is Green: 424 parser tests, Rust clippy, rebuilt WASM, workspace typecheck, 5,489 TypeScript tests, public API baseline, layout-boundary checker, and diff checks.
- GitHub CI is Green: audit, rust, smoke, and typecheck.

No unresolved review or verification findings remain.

PR #1056, opened 2026-07-19
refactor(docx): retain complete textbox stories

yukiyokotani:
## Summary

- route parser-complete `CT_TxbxContent` through the shared paragraph/table story dispatcher
- retain authored block order, nested tables, supported floating drawings, vertical writing transforms, autofit/clipping behavior, and ordered unsupported-block diagnostics
- preserve the public `ShapeRun.textBlocks` projection as an explicit compatibility fallback
- remove the legacy paint-time text acquisition/autofit measurement path and reduce the layout-boundary baseline in the same change
- keep nested floating-table geometry and source/resolved ownership aliases in the vertical text-box local frame when `spAutoFit` shifts the retained story

## Specification basis

- ECMA-376 §20.4.2.38 (`wps:txbxContent` / WordprocessingML block content)
- ECMA-376 §21.1.2.1.1 (`a:bodyPr`, insets and vertical direction)
- ECMA-376 §21.1.2.1.4 (`a:spAutoFit`)

## Review

Claude Fable 5 was consulted for the story-acquisition design and identified the vertical `spAutoFit` floating-table coordinate mismatch during targeted review. The defect was reproduced with a focused failing test, fixed with a text-box-local translation path, and reverified across the retained-layout graph.

## Verification

- `vitest run packages/docx/src` — 279 files, 2568 tests
- `pnpm test` — 534 files passed, 11 skipped; 5495 tests passed, 26 skipped
- `pnpm typecheck`
- `pnpm test:docx-boundaries` — 83 tests
- `pnpm test:docx-public-api` — 19 tests
- `pnpm --filter @silurus/ooxml-docx build`
- `git diff --check`
- GitHub CI — audit, rust, smoke/worker-equivalence, and typecheck/build/API jobs passed

Part of #1037.


yukiyokotani:
Review follow-up is now pushed in `8fb1237f`.

Claude Fable 5's targeted analysis found that vertical `spAutoFit` shifts the whole retained text-box story in a local frame, while the generic table occurrence translator intentionally leaves resolved floating-table placements in their page-owned frame. That made a nested floating table lag behind its host story.

The fix adds a vertical text-box-local table translation path, moves both source and resolved placement geometry, and memoizes source placements so `resolved.source` and `resolved.child` retain their canonical aliases. A focused regression test was observed Red before the fix and Green afterward. Full DOCX, repository, type, architecture-boundary, public API, and package-build checks pass locally.

yukiyokotani:
CI follow-up in `4dfa49bc`:

The fresh run reproduced the same unrelated failure as the previous run: the unchanged parser-backed vertical header/footer pagination probe exceeded Vitest's default 5-second per-test timeout on the constrained runner. All other 5,490 tests passed, and the focused probe completes locally in about 0.3 seconds of test time.

Because this recurred twice, the probe now has an explicit 15-second timeout. Its two-document, 40-paragraph workload and all assertions remain unchanged. The focused probe, full repository suite (5,495 passed), lint, and diff check pass locally. Fresh CI is running on the new head.

PR #1057, opened 2026-07-19
refactor(docx): make page layers own drawing order

yukiyokotani:
## Summary

- materialize anchored drawing order as immutable `PageLayers.paintOrder`
- retain plain-data transform, clip, coordinate-space, and owner facts for nested anchors
- make Canvas paint consume the completed order without graph discovery, sorting, or callback re-entry
- remove the legacy deferred front-paint session and renderer callback queue
- document normative anchor ordering separately from cross-story compatibility preservation

## Specification and compatibility

ECMA-376 Part 1 §20.4.2.3 defines `behindDoc` grouping and `relativeHeight`
ordering within an equivalent stacking context. The implementation applies that
rule independently to retained header, body, notes, and footer story runs.

The standard does not define a single total order across those stories. This PR
therefore preserves the established page-root order and records that decision in
`docs/docx-page-layer-order.md`. Synthetic cross-story fixtures are ready, but a
fresh-machine Word observation has not yet been recorded because desktop
automation permission and Word-for-web authentication are unavailable.

## Verification

- `pnpm test` — 5,495 passed, 26 skipped
- DOCX package production build and declaration generation
- TypeScript project build
- DOCX public API baseline and 19 mutation tests
- layout-boundary checker and focused callback/import mutation tests
- `ast-grep scan` — exit 0; existing repository warnings only
- `git diff --check`

## Notes

- No parser source changed, so generated WASM was not rebuilt.
- No public DOCX declaration changed.
- The boundary baseline did not require expansion; the checker now rejects the
  removed callback identifiers and import path explicitly.
- Independent Claude Fable 5 review found no blockers. Its one medium finding,
  concerning local anchors in non-owned text boxes, is fixed with a regression
  test in `596a7c42`.

Part of #1037.


yukiyokotani:
### Independent review gate

Claude Fable 5 (`claude-fable-5`, high effort) reviewed the published PR diff
against `main` without edit or GitHub-write permission.

Result: **no blocking findings**.

One medium correctness gap was identified and fixed in `596a7c42`: a text box
not owned by a materialized drawing inherited the page-root anchor suppression
flag, so descendant local anchors could be skipped when the same root contained
another page-materialized anchor. The painter now explicitly re-enters that
free-standing text-box story as a local stacking context, with a synthetic
regression test proving both local and page anchors paint once.

The reviewer also noted invariant-hardening and test-coverage opportunities.
The current integration suite already covers resolved floating tables, nested
table/cell clips, page-owned axis undo, vertical section matrices, clone safety,
and worker parity. The remaining material residual risk is the intentionally
unobserved cross-story Office order documented in
`docs/docx-page-layer-order.md`; this PR preserves existing root order rather
than claiming an unsupported compatibility rule.

Post-fix local verification:

- 5,495 tests passed; 26 skipped
- TypeScript project build
- DOCX public API and layout-boundary checks
- `git diff --check`

PR #1058, opened 2026-07-19
refactor(docx): project text geometry from layout

yukiyokotani:
## Summary

- derive search and selection geometry from retained `DocumentLayout` text placements in semantic reading order
- keep `layout/text-index.ts` entirely in physical page points and project to CSS pixels in a dedicated root adapter
- recursively include anchored text-box stories at their host while using `PageLayers.paintOrder` only for already-retained transform geometry
- preserve internal and external hyperlink targets across table, vertical-section, text-box, floating-table, and page-owned anchor transforms
- remove paint-time text callbacks, throwaway collection canvases, worker dry renders, and the obsolete vertical overlay adapter
- move pure affine algebra into an audited layout contract consumed by paint through one exact facade

## Specification and behavior

ECMA-376 Part 1 sections 17.13.6.2 and 17.16.22/17.16.23 provide bookmark and hyperlink ownership. Presentation z-order does not determine search offsets. Retained paragraph drawing order supplies the single source ordinal, avoiding a comparison between anchor run indexes and drawings-array indexes.

Footnote and endnote flow domains now use the same section-region resolution as paint: an absent `sectionRegionId` falls back to the first section region, while a dangling retained id fails loudly. This is directly covered on a rotated section.

## Review-driven correction

Claude Fable 5 reviewed the public draft and GitHub CI independently identified that the first version placed CSS display-scale projection under `layout/`. The correction does not weaken the rule: point-space traversal stays under `layout/`, CSS/font/width adaptation moved outside it, and boundary mutation tests pin the allowed affine facade and projection-adapter surface.

Direct regressions cover resolved floating-table absolute coordinates, note fallback and invalid references, mixed anchored/inline drawing order, display-scale validation, worker/main parity, and zero-Canvas collection.

## Verification

- `pnpm test`: 5,507 passed, 26 skipped
- DOCX tests: 2,580 passed
- TypeScript 6.0.3 workspace typecheck
- ast-grep scan
- DOCX layout boundary check
- 87 boundary mutation tests
- `git diff --check`

Part of #1037.


yukiyokotani:
### Independent review follow-up

Claude Fable 5 returned GO on the corrected public diff and reported two non-blocking hardening findings. Both were addressed before merge in `93ea6a03`:

- anchored and inline text-box drawings now share the paragraph run index retained in `SourceRef.path` as their single authored reading-order ordinal, with stable encounter order only within the same run;
- the audited paint affine facade now rejects aliased exports, including same-signature helper swaps, with a direct mutation regression.

Fable re-audited the drawing producers and projection paths and returned final GO. Local verification after the fixes remains green: 5,507 tests passed (26 skipped), TypeScript 6.0.3 workspace typecheck, ast-grep scan/rule tests, public API checks, DOCX package-build ownership, layout-boundary check, 87 boundary mutation tests, and diff check. Final GitHub Actions passed audit, Rust, smoke, and typecheck.

PR #1060, opened 2026-07-20
refactor(docx): unify float placement callers

yukiyokotani:
## Summary

Completes Series C1b of #1037.

- replace the transitional scalar float-overlap adapter with typed placement participants
- require floating-table overlap intent at every retained transport boundary
- apply ECMA-376 §17.4.56 `tblOverlap` as an intrinsic pairwise table constraint, regardless of source order
- keep Word-compatible cross-paragraph displacement, ordinary-flow table admission, and page-anchor deferral as separately named policies
- migrate the remaining floating-table parent/child and page-anchor retry loops to exact-state convergence with explicit non-convergence failures
- harden the DOCX layout boundary checker against namespace/default/re-export/dynamic/CommonJS/property-access bypasses

## Verification

- DOCX source suite: 2,609 tests passed
- full repository suite with bounded thread pool: 5,536 tests passed, 26 skipped
- boundary mutation suite: 90 passed
- `pnpm typecheck`
- `pnpm lint`
- `pnpm lint:test`
- DOCX production/declaration build
- DOCX public API declaration baseline
- `git diff --check`

## Design and review

Claude Fable 5 reviewed the policy split and fixed-point migration before implementation. Its first public-diff review found two blockers (compatibility right-boundary semantics and a redundant stable convergence pass); both were fixed with regression coverage. The second public-diff review approved the updated PR with no remaining blockers.

### DIFF
--- a/docs/docx-layout-engine-series-c-plan.md
+++ b/docs/docx-layout-engine-series-c-plan.md
@@ -34,6 +34,7 @@
 - Modify: `packages/docx/src/layout/compatibility.test.ts`
 - Modify: `packages/docx/src/layout/convergence.ts`
 - Modify: `packages/docx/src/layout/line-wrap-convergence.ts`
+- Modify: `packages/docx/src/layout/body-paginator.ts`
 - Modify: `packages/docx/src/layout/paragraph.ts`
 - Modify: `packages/docx/src/layout/floating-table-transaction.ts`
 - Modify: `packages/docx/src/layout/table-pagination.ts`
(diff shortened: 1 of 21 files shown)


yukiyokotani:
Fable review round 1 requested two changes; both are addressed in 26373e70:

1. Restored the compatibility-path right-boundary contract by translating the exclusion right padding into raw-object space. Added policy-level boundary/slack coverage and a display-transport regression.
2. Removed the redundant full floating-table fragment-layout pass in the stable common case. The exact-state confirmer now returns the prior candidate before layout and reuses its cached fingerprint.

Also fixed the non-blocking frame participant kind and added a mixed normative/compatibility attribution test.

Post-fix verification:
- DOCX source suite: 2,609 passed
- full repository bounded-thread suite: 5,536 passed, 26 skipped
- boundary mutation suite: 90 passed
- typecheck, lint, lint:test, DOCX build, public API baseline, and diff check pass

Requesting Fable round 2 against the updated public diff.

yukiyokotani:
Claude Fable 5 round 2 result: **APPROVE**.

It re-derived the exclusion-space/object-space boundary identity, confirmed the exact slack tie and display-path regression, verified the stable transaction shortcut preserves deterministic exact-state convergence and cycle detection, and checked compatibility-rule attribution plus the frame-kind correction. No actionable blockers remain.

PR #1068, opened 2026-07-21
refactor(docx): narrow acquisition capabilities

yukiyokotani:
## Summary

- replace body acquisition's structural Canvas dependency with a runtime text-measurement facade exposing only font selection, spacing, kerning, and `measureText`
- inject DOM-aware vertical-glyph measurement as a separate private service-owner capability, include it in text fingerprints, and fail closed for unbound vertical layout
- compose one frozen, identity-preserving parser projection record for acquisition facts instead of adding parser-model back-edges or compatibility wrappers
- extend the DOCX architecture ratchet to require these seams and reject paint capabilities, renderer back-edges, missing modules, and declaration expansion

This is the C3d-alpha capability-seam step for #1037. It intentionally prepares the physical acquisition-cluster relocation planned for C3d-beta; it does not move those clusters yet or change OOXML semantics.

## Design notes

The measurement facade delegates to the exact same concrete context used by the vertical-glyph service, so existing metrics and kerning state remain coherent while acquisition receives no backing canvas or painter. DOM vertical probing and worker-safe measurement have distinct fingerprints to prevent incompatible retained layouts from sharing a key.

The parser projection record preserves the identities of the existing normalization functions. It is a required acquisition capability, not a wrapper or sample-specific heuristic.

## Verification

- full repository Vitest: 5,616 passed, 26 skipped
- DOCX Vitest: 2,689 passed
- TypeScript workspace typecheck
- DOCX boundary checker and 93 mutation tests
- public API baseline and 19 mutation tests
- compatibility evidence (65 rules) and 17 mutation tests
- DOCX package-build ownership test
- ast-grep scan and 8 lint-rule tests
- `git diff --check`

### DIFF
--- a/packages/docx/src/body-layout-context-integration.test.ts
+++ b/packages/docx/src/body-layout-context-integration.test.ts
@@ -8,6 +8,7 @@ import {
   computeColumns,
   resolveBodyParagraphLayoutContext,
 } from './renderer.js';
+import { bodyAcquisitionInputProjections } from './parser-model.js';
 import type {
   DocParagraph,
   DocxDocumentModel,
@@ -58,7 +59,7 @@ describe('body layout context integration', () => {
     const layoutSettings = resolveDocumentLayoutSettings(document);
     const sectionLayout = resolveSectionLayoutContext(layoutSettings, document.section);
     const context = resolveBodyParagraphLayoutContext(
-      { layoutSettings, sectionLayout },
+      { layoutSettings, sectionLayout, acquisitionInputs: bodyAcquisitionInputProjections },
       paragraph(),
     );
 
--- a/packages/docx/src/column-widths.test.ts
+++ b/packages/docx/src/column-widths.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect } from 'vitest';
 import { createLayoutServices, resolveColumnWidths } from './renderer.js';
 import { buildSegments, layoutLines } from './line-layout.js';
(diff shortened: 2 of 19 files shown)


yukiyokotani:
Claude Fable 5 completed an independent review of the full public diff and returned **APPROVE** (no P1 findings).

The review identified one non-blocking P2: the new measurement-surface ratchet was still a denylist, so an unlisted Canvas method, a heritage clause, or an extra declaration could widen the seam without failing. Follow-up commit `6710c677` closes that gap before merge:

- exact declaration and member allowlists now pin both measurement interfaces and the parser projection record
- interface heritage, extra declarations, duplicate members, and unlisted members are rejected
- mutation cases cover `fillRect`, `extends CanvasRenderingContext2D`, and an extra escape interface
- production coverage pins the vertical service to the context/font selected by layout
- rescaling and null-context vertical measurement now fail closed

The review's remaining P3 notes are intentionally staged: the four kernel-local parser calls move with their owning clusters in C3d-beta, and the long positional `layoutLines` API is scheduled for C3e consolidation.

Post-review verification at the updated head:

- repository Vitest: 5,619 passed, 26 skipped
- TypeScript workspace typecheck
- boundary mutations: 93 passed
- public API mutations: 19 passed
- compatibility mutations: 17 passed
- DOCX package-build ownership: passed
- lint rule tests: 8 passed
- `git diff --check`

PR #1070, opened 2026-07-21
refactor(docx): own page anchor classification

yukiyokotani:
## Summary

- move page-owned versus paragraph-local DrawingML anchor classification from the renderer into the point-space layout subsystem
- remove the renderer-only classifier test adapter and test the layout owner directly
- require the new owner in the architecture boundary ratchet and reject renderer back-edges
- intentionally keep display-scale anchor/frame geometry outside layout until its transitional compatibility bridge is deleted

This is the C3d beta-2 anchor/frame ownership decision for #1037.

## Design review

GPT-5.6 Sol rejected an initial broader relocation because the legacy anchor/frame helpers consume pixel scale and Canvas-oriented geometry. Moving those helpers under layout would violate the repository's point-space boundary. The final diff relocates only the scale-independent classification rule; the pixel bridge stays isolated for removal in C3e. A follow-up review also replaced a root compatibility-facade dependency with the canonical layout/float-wrap owner. No unresolved P1/P2 finding remains.

## Specification basis

Classification preserves ECMA-376 section 20.4.3.5 positionV relativeFrom behavior and the registered Word page-level float prescan compatibility rule. No geometry or rendering behavior changes.

## Verification

- pnpm typecheck
- pnpm test: 5,633 passed, 26 skipped
- DOCX layout boundary mutation suite: 93 passed
- DOCX public API suite: 19 passed
- DOCX compatibility-evidence suite: 17 passed
- DOCX package-build ownership suite: 1 passed
- pnpm lint
- pnpm lint:test
- git diff --check


yukiyokotani:
GPT-5.6 Sol CI follow-up review found that the new classifier JSDoc described an existing Word compatibility policy outside its registered compatibility authority. Commit 66247504 narrows the comment to the ECMA-376 coordinate classification only; runtime behavior is unchanged. Reverified locally: compatibility evidence checker and 17 mutation tests, 21 focused anchor tests, typecheck, lint, and diff check all pass.

PR #1071, opened 2026-07-21
refactor(docx): inject body acquisition projections

yukiyokotani:
## Summary\n\n- route the four deferred body-kernel paragraph/table parser projections through each acquisition state's immutable capability record\n- remove direct `paragraphAcquisitionInput` and `tableFormatInput` imports from the renderer\n- add an architecture ratchet that rejects direct, aliased, namespace, CommonJS, dynamic, computed, and global-record bypasses\n\nThis is C3d beta-3 of #1037. It closes the parser-projection follow-up recorded during C3d-alpha without moving display-scale/Canvas bridges into the point-space layout layer.\n\n## Design notes\n\n`bodyAcquisitionInputProjections` retains the exact parser-owned function identities, but the concrete body kernel may access them only through `BodyAcquisitionState.acquisitionInputs`. The global record is restricted to `buildMeasureState`, where the immutable capability is attached. No OOXML behavior or public declaration changes.\n\n## Verification\n\n- `pnpm test`: 5,633 passed, 26 skipped\n- `pnpm typecheck`\n- DOCX layout boundary mutations: 94 passed\n- DOCX compatibility mutations: 17 passed; 65 rules verified\n- DOCX public API mutations: 19 passed\n- DOCX package-build ownership\n- `pnpm lint`\n- `pnpm lint:test`: 8 passed\n- `git diff --check`\n\n## Review\n\nGPT-5.6 Sol reviewed the implementation and strengthened the first boundary design to reject global projection-record, alias, computed-member, CommonJS, and dynamic-import bypasses. No unresolved P1/P2 finding remains.

### DIFF
--- a/packages/docx/src/renderer.ts
+++ b/packages/docx/src/renderer.ts
@@ -184,7 +184,6 @@ import {
   bodyAcquisitionInputProjections,
   internalDocumentModel,
   numberingMarkerShapeInput,
-  paragraphAcquisitionInput,
   publicAnchorBridge,
 } from './parser-model.js';
 import {
@@ -222,7 +221,6 @@ import {
   applyNumberingBodyOffset,
   resolveNumberingMarkerGeometry,
 } from './layout/numbering-marker.js';
-import { tableFormatInput } from './parser-model.js';
 import { resolveTableColumnWidths } from './layout/table-columns.js';
 import {
   measureParagraphIntrinsicWidths,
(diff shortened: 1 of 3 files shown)


yukiyokotani:
GPT-5.6 Sol public-diff review found one boundary-ratchet gap after the initial beta-3 implementation: the checker identified the parser-projection owner only by the enclosing function name, so a nested function named `buildMeasureState` could impersonate the real owner.

Commit `8a74efbd` closes that path by comparing the nearest function node with the actual top-level `buildMeasureState` declaration and by rejecting direct calls through the global projection record even inside the owner. A focused mutation test now pins the same-name nested-function bypass.

Post-review verification:
- DOCX layout boundary mutation suite: 94/94 passed
- transitional repository boundary check: passed
- lint: passed (pre-existing warnings only)
- lint rule tests: 8/8 passed
- `git diff --check`: passed

The explicit `--final` check still reports the transitional paragraph-anchor-frame adapter, which is intentionally scheduled for Series C3e rather than hidden in this beta-3 ownership move.

PR #1078, opened 2026-07-22
feat(pptx): expose shape id on text runs

z2014:
## Summary

- expose the source shape's slide-local `cNvPr@id` as optional `PptxTextRunInfo.shapeId`
- propagate the identifier to every text run emitted for a file-authored shape
- keep the field absent for parser-synthesized shapes without `cNvPr`

## Motivation

Consumers of Canvas text-run geometry need a stable source-shape handle for selection and edit integrations. The PPTX parser already retains `cNvPr@id` on `ShapeElement`, but the renderer previously dropped that identity from text-run callbacks.

## Validation

- `vitest run packages/pptx/src/smartart-fallback-contrast.test.ts` (16 tests)
- `vitest run packages/pptx/src` (477 tests)
- `pnpm --filter @silurus/ooxml-pptx build:ci`
- `pnpm build`
- `pnpm lint` (passes with existing warnings)

### DIFF
--- a/packages/pptx/src/renderer.ts
+++ b/packages/pptx/src/renderer.ts
@@ -136,6 +136,11 @@ export interface RenderContext {
 
 /** Information about a rendered text segment for building a transparent selection overlay. */
 export interface PptxTextRunInfo {
+  /**
+   * Source shape's DrawingML `cNvPr@id`. The identifier is stable and unique
+   * within its slide. Absent for parser-synthesized shapes that have no cNvPr.
+   */
+  shapeId?: string;
   text: string;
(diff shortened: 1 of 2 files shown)


z2014:
增加shapeid字段，就能和officecli工具串起来了，review下看看

yukiyokotani:
Thank you for proposing this integration. Exposing the source `cNvPr@id` is a clean way to connect rendered text geometry back to OfficeCLI's stable shape addressing, and the focused tests made the change easy to validate. The PR has been merged, and I’m following up with the corresponding DOCX/XLSX identities where those formats provide an appropriate stable address.

Issue #496, opened 2026-06-17
Is editing on the roadmap, or out of scope?


evrenverse:
Thanks for this project — the rendering fidelity is impressive, and covering docx/xlsx/pptx in one library is great.

From the docs and the `.d.ts`, the viewers look read-only by design (canvas rendering, with text/cell selection + copy, but no mutation API).

Two questions:

1. Is any form of **editing** — even narrow, e.g. filling form fields / content controls in `.docx`, or editing cell values in `.xlsx` — on the roadmap or under consideration?
2. Or is editing considered fundamentally **out of scope** here — i.e. a separate, much more complex product (a full OOXML editor with edit→layout round-trip), rather than something this viewer would grow into?

Just trying to understand the long-term direction. Thanks!

yukiyokotani:
Thanks for the kind words, and for such a well-framed question — you've read it exactly right: the viewers are read-only by design (canvas rendering, with text/cell selection + copy, but no mutation API).

Short answer: editing — including the narrow forms you mention (form fields, content controls, cell values) — is out of scope for now. That's a deliberate stance, not just "haven't gotten to it yet." Three reasons:

1. **Rendering has a well-defined notion of "correct"; editing doesn't.** "Given this OOXML, what should it look like" is something we can hold to a high bar. Editing is a stateful operation whose only real oracle is "what Word writes back" — expensive to pin down, and noisy, since Word/Excel/PowerPoint normalize and rewrite parts of the file on save (even a no-op edit produces a diff). The property that makes this library trustworthy for viewing is exactly the one that's missing for editing.

2. **"Narrow" editing isn't actually that narrow, and the line is hard to draw.** Once any mutation ships, the boundary between "this edit is supported" and "this edit silently does the wrong thing" becomes the project's problem — and that boundary is precisely where a fidelity-focused viewer would lose trust. Editing an **xlsx cell value** meaningfully, for instance, implies a formula recalculation engine: change one cell and every dependent cell's cached result goes stale, so without recalc you've produced an internally inconsistent workbook. A half-supported editor would undercut the one thing the library is currently good at.

[...]

PR #1081, opened 2026-07-23
feat(pptx): support in-memory shape updates

z2014:
## Summary

- add atomic `PptxPresentation.applyShapeChanges()` mutations for slide-owned shapes
- expose one serializable `PptxShapeChangeType.Update` operation with a typed top-level ShapeElement delta
- replace nested values such as `textBody` as complete values instead of exposing paragraph/run operations
- return detached `applied` and correctly ordered `inverse` batches for request adapters and undo/redo
- return the affected slide so callers can redraw only that slide

## Motivation

Canvas consumers can now correlate rendered text with a slide-local shape ID, but the parsed presentation model was private and read-only. Editors therefore had to reload and reparse the whole PPTX after every change. This API lets a caller optimistically update one shape in memory and redraw only the affected slide.

## API

```ts
import { PptxShapeChangeType } from '@silurus/ooxml/pptx';

const result = presentation.applyShapeChanges({
  slideIndex,
  shapeId,
  changes: [
    {
      type: PptxShapeChangeType.Update,
      patch: {
        x: 914400,
        fill: { fillType: 'solid', color: '4472C4' },
        textBody: updatedTextBody,
      },
    },
  ],
});

await presentation.renderSlide(canvas, result.slideIndex, { width: 960 });

// `result.applied` is suitable for logging or request adaptation.
// `result.inverse` can be passed back directly to undo the batch.
```

Each batch runs on an isolated structured clone. Patches are shallow: omitted top-level properties stay unchanged, while supplied nested values are replaced whole. Invalid targets, unsupported non-JSON values, or `type` / `id` changes leave the live model unchanged. Optional properties can be removed with `unset`, which is also used by generated inverse deltas. The API is main-mode only; worker mode keeps its model inside the render worker.

## Validation

- `vitest run packages/pptx/src` (73 files, 486 tests)
- `pnpm --filter @silurus/ooxml-pptx build:ci`
- `pnpm lint` (passes with existing warnings)
- `git diff --check`


z2014:
support update model api

z2014:
@yukiyokotani cc

yukiyokotani:
Thank you for the thoughtful implementation. I can see a lot of care went into the design, including the atomicity guarantees, undo/redo support, and comprehensive tests.

That said, I don't think I can merge this PR—not because of the implementation quality, but because of the project's scope.

As I mentioned in #496, the design goal of this library is not simply to avoid providing an editing UI, but to remain a **read-only viewer**.

Introducing a public mutation API such as `applyShapeChanges()` would significantly expand the responsibility of the project.

Although the current scope is limited to top-level shape properties, once such an API exists, it naturally leads to expectations for:

* more shape types
* insert/delete operations
* text-level editing
* worker mode support
* equivalent APIs for DOCX/XLSX
* persistence

In particular, XLSX illustrates how difficult it becomes to define the boundary. Even changing a single cell value raises questions about formula recalculation, cached values of dependent cells, shared formulas, tables, defined names, pivot tables, and many other aspects of workbook consistency. What appears to be a simple edit can easily produce an internally inconsistent workbook.

As a result, exposing only a subset of editing functionality makes it very difficult to define what is and isn't safely supported. Even worse, unsupported operations may appear to work while silently producing invalid document state. I don't think it's realistic to keep those boundaries clear as part of this project's responsibilities.

The parser and in-memory model are also intentionally designed to retain only the information required for rendering. They are not intended to evolve into a general-purpose editing model.

I think the implementation itself is very well done, and I completely understand the use case. However, it doesn't align with the long-term direction of this repository, so I'm going to pass on merging it.

That said, this decision is not a criticism of the idea itself. Since this is an open-source project, I'm completely happy for someone to fork it and evolve it into an editing-focused library, or even take it in an entirely different direction. In fact, I'd welcome that. If such a project eventually grows into something that surpasses this library for editing use cases, I think that would be a great outcome for the open-source ecosystem.

Thank you again for the time and effort you put into this contribution.

PR #1085, opened 2026-07-23
feat(pptx): add shape click hit testing

z2014:
## Summary

- add main-mode `PptxPresentation.hitTestShape()` model-backed shape hit testing
- add `PptxViewer.onShapeClick` with slide-local shape IDs and detached shape snapshots
- account for rotation, flips, reverse paint order, and configurable straight-line tolerance
- keep hit testing independent of renderer callbacks and pixel output

## API

```ts
const viewer = new PptxViewer(canvas, {
  onShapeClick({ slideIndex, shapeId, shape, point, nativeEvent }) {
    console.log(slideIndex, shapeId, shape, point, nativeEvent);
  },
});
```

Headless canvas integrations can convert client coordinates to slide EMU and
call the same model-backed hit test:

```ts
const hit = presentation.hitTestShape(slideIndex, point, {
  tolerance: lineToleranceInEmu,
});
```

## Behavior

- walks slide elements in reverse paint order and returns the topmost identified `ShapeElement`
- inverse-transforms the point for rotated and flipped shape frames
- uses segment distance plus caller-provided tolerance for straight lines and connectors
- ignores parser-synthesized shapes without a stable DrawingML `cNvPr@id`
- returns a detached shape snapshot instead of exposing the live presentation model
- listens on the viewer wrapper so selectable text-overlay clicks still reach shape hit testing
- ignores prevented hyperlink clicks
- requires `mode: "main"` because worker mode does not retain the parsed model on the viewer thread

## Scope

This PR does not modify the PPTX renderer or introduce shape mutation APIs.

## Validation

- `vitest run packages/pptx/src` (75 files, 490 tests)
- PPTX Vite production and declaration builds
- `ast-grep scan` (passes with existing warnings)
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,39 @@ Notes:
   transferred back as an `ImageBitmap`, so a single render can be marginally
   slower than `mode: 'main'`. Choose it for non-blocking UI, not raw speed.
 
+### PPTX shape click and hit testing
+
+In `mode: 'main'`, `PptxViewer` can report the topmost identified shape under a
+click. The event includes the slide-local DrawingML id and a detached
+`ShapeElement` snapshot without exposing the presentation's private model.
+
+```typescript
+const viewer = new PptxViewer(canvas, {
+  onShapeClick({ slideIndex, shapeId, shape, nativeEvent }) {
+    console.log(slideIndex, shapeId, shape, nativeEvent);
+  },
+});
+```
+
+Headless integrations can call the same model-backed hit test directly. Points
+and tolerance use slide EMU coordinates:
+
+```typescript
+const rect = canvas.getBoundingClientRect();
+const point = {
+  x: ((event.clientX - rect.left) / rect.width) * pres.slideWidth,
+  y: ((event.clientY - rect.top) / rect.height) * pres.slideHeight,
+};
+const hit = pres.hitTestShape(slideIndex, point, {
+  tolerance: (6 / rect.width) * pres.slideWidth,
+});
+```
+
+Hit testing walks shapes in reverse paint order, uses each shape's transformed
+frame, and applies the optional tolerance to straight lines and connectors.
+Shapes without a file-authored id are ignored. Both `onShapeClick` and
+`hitTestShape()` require `mode: 'main'`.
+
 ### Continuous scroll viewers
 
 `DocxScrollViewer` and `PptxScrollViewer` render the whole document as one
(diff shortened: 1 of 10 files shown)


z2014:
@yukiyokotani cc

yukiyokotani:
@z2014 
Thanks for the PR.

Before reviewing the implementation itself, I'd like to better understand the long-term goal behind it.

Is your end goal to support editing, with this PR being one piece toward that?

On its own, I don't have any particular concerns with adding shape hit testing. However, if the intended direction is to eventually support editing (for example through MCP or another external integration), then I think we end up at the same discussion as in #1081.

As I mentioned there, I don't plan for this project to grow into an editing foundation or expose the kinds of APIs that would naturally lead in that direction. So if this PR is primarily a stepping stone toward that goal, I don't think it can ultimately reach the destination you're aiming for within this repository.

I'd appreciate hearing a bit more about the intended use case before I review the implementation in detail.

yukiyokotani:
Thank you for proposing this interaction. We revisited it while redesigning the Viewer selection API after #1176, particularly for read-only AI/MCP workflows where a user clicks the content they want to ask about.

#1181 incorporates that Viewer-side use case through `enableElementSelection`, `onSelectionContextChange`, and `getSelectionContext()`. It hit-tests the topmost transformed element in both main and worker modes and returns a compact, detached snapshot with bounds, provenance, and bounded descriptive content.

The new API intentionally does not return the live `ShapeElement` or native DOM event. Those values couple consumers to parser/model lifetime and an editor-oriented object surface, while the Viewer needs a serializable read-only focus boundary that is safe to hand to another process. Event-specific UI data remains available to applications from their own DOM listener when needed.

I am leaving this PR open until #1181 completes its independent merge-gate review. Once that lands, I plan to close this as superseded, with the useful interaction requirement preserved in the unified API. Thanks again for identifying it.

yukiyokotani:
Thank you again for proposing PPTX shape hit testing. The read-only Viewer use case behind this PR has now been incorporated in #1181, which exposes bounded, detached selection context through `getSelectionContext()` and `onSelectionContextChange`.

#1181 includes reverse-paint-order hit testing, transformed frame and straight-line tolerance, main-thread/worker parity, and stable element provenance. It deliberately does not expose editing operations or native DOM events.

I’m closing this PR as superseded by #1181. This proposal was an important input to the design and is credited in the follow-up. If there is Viewer-only behavior that the merged API does not cover, please open a focused issue or PR and we can evaluate it against the new selection-context contract.

PR #1119, opened 2026-07-30
feat: bound aggregate OOXML archive inflation

the-laughing-monkey:
## Summary

- add a shared fail-closed ZIP budget covering per-entry bytes, aggregate expanded bytes, and archive entry count
- enforce declared archive limits before reads and actual post-inflate bytes while entries are consumed
- retain one budget across DOCX, XLSX, and PPTX parser objects and propagate limits through main/render worker protocols
- preserve poisoning after a proven breach so later reads cannot continue from a partially trusted archive

## Verification

- `cargo test --workspace`
- `pnpm exec vitest run packages/docx/src packages/pptx/src packages/xlsx/src` (431 files, 3,982 tests)
- `pnpm typecheck`
- `pnpm build`
- `pnpm lint` (pre-existing warning-only tripwires)

## Notes

The full root Vitest command also exercises unrelated Node pixel-equality tests; two effect-crop assertions failed and the run did not terminate promptly. All DOCX/XLSX/PPTX package suites and the complete Rust workspace are green.

This PR does not change rendering behavior or add a compatibility fallback.

### DIFF
--- a/packages/docx/parser/src/lib.rs
+++ b/packages/docx/parser/src/lib.rs
@@ -1,3 +1,4 @@
+use std::io::Cursor;
 use wasm_bindgen::prelude::*;
 
 mod drawing_compatibility;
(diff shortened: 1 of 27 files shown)


yukiyokotani:
Thank you for this PR. Its retained-session budget and poisoning model directly influenced the resource-governance design released in v0.75.0 through #1120 and #1130.

### Incorporated

- early per-entry checks against declared size, followed by authoritative actual-byte checks during inflation;
- one budget retained across lazy reads in a package session;
- aggregate accounting across distinct entries actually read;
- poisoning after a proven limit violation;
- propagation through browser main/render workers and Node sessions; and
- a bounded central-directory preflight with an internal archive-entry-count ceiling.

These now use one shared policy and typed `OoxmlResourceLimitError` contract across DOCX, XLSX, and PPTX.

### Changed in the merged design

- The public aggregate budget counts actual bytes from distinct entries that are read. It does not reject from the archive-wide sum of untrusted declared sizes or charge the same entry repeatedly.
- The public defaults are 128 MiB per entry and 256 MiB distinct total. Separate internal hard ceilings remain active even when a configurable byte limit is `null`.
- The policy is an admission budget, not an exact memory quota: compressed size and ZIP declarations cannot predict XML/model/JSON/WASM/Canvas/browser amplification.

### Not fully addressed

- v0.75.0 enforces an internal hard ceiling of 20,000 archive entries before the ZIP library builds its owned index.
- Applications cannot yet choose a stricter entry-count policy. This is a design gap, not a technical impossibility; `archiveEntryCount` is already available in metrics.
- Public configurability is tracked in [#1133](https://github.com/yukiyokotani/office-open-xml-viewer/issues/1133). Feedback there on the contract and defaults would be welcome.

The current migration and policy documentation is in the [v0.75.0 release](https://github.com/yukiyokotani/office-open-xml-viewer/releases/tag/v0.75.0). Since the merged implementation incorporates the proposal's main invariants in the shared cross-format architecture, we are closing this PR as superseded. Thank you again for the contribution.

PR #1124, opened 2026-07-31
fix: terminate XLSX and PPTX workers after rejected loads

the-laughing-monkey:
## Summary

- clean up partially initialized XLSX and PPTX instances when loading rejects
- terminate the worker directly if construction fails before an instance exists
- add regression coverage for both rejected-load paths

## Why

DOCX already destroys its partially initialized engine when loading fails, but XLSX and PPTX did not. A parse/load rejection could therefore leave the worker and its bridge alive until the page itself was torn down.

This changes only failure cleanup. Successful loads and the public API are unchanged.

## Validation

- PPTX and XLSX package suites: 137 files, 1,083 tests passed
- `pnpm typecheck`
- `pnpm build`
- `pnpm lint` (existing warning-only tripwires)

### DIFF
--- a/packages/pptx/src/presentation-destroy.test.ts
+++ b/packages/pptx/src/presentation-destroy.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, afterEach } from 'vitest';
+import { describe, it, expect, vi, afterEach } from 'vitest';
 import { WorkerBridge, preloadGoogleFonts, type WorkerLike, type FontPreloadEntry } from '@silurus/ooxml-core';
 import { PptxPresentation } from './presentation';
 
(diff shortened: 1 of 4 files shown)


yukiyokotani:
Thank you for identifying and testing this cleanup path. The fix is included in v0.75.0 through #1120 and #1130, using the same failure-cleanup rule for DOCX, XLSX, and PPTX.

### Released behavior

- If loading fails after an engine has been partially constructed, the engine is destroyed.
- If construction fails before an engine exists, the underlying worker is terminated directly.
- Cleanup does not replace the original load error.
- The shared path is covered by cross-format regression tests.

Successful-load behavior and the existing Viewer `load()` interfaces remain compatible. The shared ownership work also closes late transfers, cancels pull sessions, releases cached/decoded resources, and prevents a rejected load from leaving a live worker bridge behind.

Because this fix is now part of the merged shared lifecycle implementation, we are closing this PR as superseded. If the original leak can still be reproduced on [v0.75.0](https://github.com/yukiyokotani/office-open-xml-viewer/releases/tag/v0.75.0), please let us know.

PR #1140, opened 2026-08-03
feat(pptx): expose element provenance

z2014:
## Summary

- expose a parallel `elementSources` array for parsed slide elements
- distinguish master, layout, and direct slide origins
- record native slide-tree indexes for one-to-one direct slide shapes

## Why

The PPTX parser currently flattens inherited and direct shapes into one rendering array. Consumers cannot reliably distinguish the source part or translate the composite array position back to the direct slide shape tree.

## Validation

- `cargo test -p pptx-parser element_sources_distinguish_inherited_and_direct_shapes`
- `cargo test -p pptx-parser master_sptree`
- `pnpm exec tsc --build packages/pptx/tsconfig.json --pretty false`
- rebuilt the PPTX WASM parser and verified parser output against a public fixture


z2014:
@yukiyokotani cc

yukiyokotani:
Thanks for the PR.

Before reviewing the implementation, please explain the concrete read-only viewer or rendering use case that requires mapping rendered elements back to their native slide-tree positions.

As explained in #1081, this project intentionally keeps its parser and in-memory model limited to information required for read-only rendering; it is not intended to become an editing or general-purpose document-analysis foundation. I also asked in #1085 whether the related API work was intended as a step toward editing, but have not yet received clarification.

As currently described, `elementSources` and `slideTreeIndex` appear primarily intended to support editing or external analysis rather than rendering. I have therefore marked this PR as `out of scope` for now. I’ll leave it open so you can clarify the intended viewer use case.

yukiyokotani:
Thank you for raising element provenance. We reconsidered it as part of the unified read-only selection-context work, because an assistant explaining clicked slide content benefits from knowing whether it came from the master, layout, or slide itself.

#1181 retains one provenance entry per rendered element and returns `master | layout | slide` in compact element context, in both main and worker modes. The parser test covers a composite paint list containing all three origins.

The design intentionally omits `slideTreeIndex`. A rendered slide is a composite of inherited and slide-authored content and can also contain synthesized fallback elements, so that index would look like a stable editor identity without being a lossless round-trip coordinate. `elementIndex` is instead documented as snapshot-local paint order; provenance supplies the Viewer-relevant explanation without promising mutation semantics.

I am leaving this PR open until #1181 completes its independent merge-gate review. Once that lands, I plan to close this as superseded. Thanks again—the provenance requirement is preserved, with its semantics made explicit.

yukiyokotani:
Thank you again for proposing PPTX element provenance. The read-only Viewer use case behind this PR has now been incorporated in #1181 as part of the unified selection-context API.

The merged implementation exposes bounded, detached element context through `getSelectionContext()` and `onSelectionContextChange`, preserves slide/layout/master provenance in both main-thread and worker modes, and keeps the API focused on inspection rather than editing.

I’m closing this PR as superseded by #1181. This proposal materially informed the provenance design and is credited in the follow-up. If there is Viewer-only provenance information that the merged contract does not cover, please open a focused issue or PR and we can extend the discriminated context model without introducing a parallel interface.

PR #1181, opened 2026-08-08
feat(viewers): expose bounded selection context

yukiyokotani:
## Summary

- Add one `getSelectionContext()` query across DOCX, XLSX, and PPTX viewers.
- Return detached, JSON-serializable snapshots discriminated by `{ format, kind }`.
- Surface DOCX/PPTX native text selection with source locators, XLSX canonical range context, and PPTX topmost element click focus.
- Support PPTX element hit testing in both main and worker modes, including master/layout/slide provenance.

## Design rationale

The public model represents the user's current focus, not an operation to perform. This avoids parallel AI-, MCP-, shape-click-, and editor-oriented interfaces for the same use case. It also leaves future read-only focus targets extensible through the existing discriminated family.

The API deliberately exposes no mutation commands, native DOM events, live parser/renderer objects, archive paths, save handles, or editor tree positions. PPTX `elementIndex` is only the paint-order index in the current rendered snapshot. Provenance follows the composite master/layout/slide paint model described by ECMA-376 sections 19.3.1.38 and 19.3.1.39.

This incorporates the Viewer-useful motivation from #1085 and #1140 while keeping editing and round-trip concerns out of scope. Thanks to @z2014 for raising those interaction and provenance use cases.

## Resource and privacy boundary

- Native selected text: at most 65,536 UTF-16 code units, without splitting a surrogate pair.
- Native run locators: at most 1,024, scanning only mounted tagged text surfaces.
- XLSX: populated cells only, with independent hard cell/text caps; no dense expansion of whole-row, whole-column, whole-sheet, or sparse multi-area selections.
- PPTX: reverse paint-order frame hit testing; only the hit element's descriptive text is streamed into a bounded result.
- Every result reports truncation, and no Viewer sends context over the network.

The detailed contract and extension policy are documented in `docs/selection-context.md`; README and site API/capability references are updated as well.

## Verification

- Affected TypeScript suites: 392 files / 3,717 tests passed.
- PPTX parser: 224 tests passed.
- `cargo clippy -p pptx-parser --all-targets -- -D warnings`
- TypeScript 6 project build.
- Root Vite production build and declaration bundle.
- Site Astro production build.
- DOCX/XLSX/PPTX public API baselines, Viewer API symmetry, and public type-export checks.
- `git diff --check`.


yukiyokotani:
## Independent adversarial merge-gate review

An independent agent reviewed the final implementation as a read-only merge gate, including public API semantics, callback-free getter use, privacy boundaries, large-input behavior, parser provenance, main/worker parity, lifecycle races, documentation, and migration compatibility.

The first pass found concrete issues rather than rubber-stamping the design. The final commits resolve and regression-test each one:

- Native text is now assembled only from selected offsets of tagged runs, bounded to the public limit plus two UTF-16 look-ahead units. `Selection.toString()` is never materialized, and untagged inter-surface DOM content cannot enter context.
- Text focus remains authoritative even without a callback and permanently retires prior/pending element focus.
- Successful reload, desk/outside click, slide navigation, slide ABA, and hidden-slide redirect all advance the element-focus generation; failed atomic reload preserves the old deck and focus.
- Consumer callback failures run after load/navigation commit and rendering, outside engine error routing.
- Old-worker hit promises are invalidated in the resource swap's before-commit phase, before worker destruction can reject them.
- Site wording now matches the implemented straight-line tolerance contract.

The reviewer rechecked head `6401410f68f908ff70be249d9db349bebd1a172c`, ran focused tests and TypeScript 6, compared the broader 392-file / 3,717-test result, and reported **Approve — no remaining P0, P1, or P2 findings**.

PR #1309, opened 2026-08-19
feat(pptx): expose ids for non-shape elements

z2014:
## Summary

- expose optional OOXML `cNvPr` IDs on picture, table, chart, and media elements
- preserve the source ID for picture-like elements emitted from slide drawing nodes
- update the public PPTX type declaration baseline

## Why

Only shape elements surfaced their stable drawing ID. Picture, table, chart, and media parsing discarded the equivalent `cNvPr` identity, so consumers could not construct stable references for those elements.

## Implementation

- picture and media IDs come from `p:nvPicPr/p:cNvPr`
- table and chart IDs come from `p:nvGraphicFramePr/p:cNvPr`
- missing IDs remain omitted from serialized output for backward compatibility

## Verification

- `cargo test -p pptx-parser --no-default-features` (262 passed)
- `pnpm exec vitest run packages/pptx/src` (669 passed)
- `pnpm --filter @silurus/ooxml-pptx build`
- `git diff --check`

### DIFF
--- a/packages/pptx/api/public-api-baseline.d.ts
+++ b/packages/pptx/api/public-api-baseline.d.ts
@@ -156,6 +156,7 @@ export interface ChartDisplayUnitsLabel {
 }
 export interface ChartElement {
     type: 'chart';
+    id?: string;
     x: number;
     y: number;
     width: number;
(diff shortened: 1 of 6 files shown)


z2014:
@yukiyokotani cc

z2014:
@yukiyokotani cc

PR #1316, opened 2026-08-20
feat(pptx): expose text box marker on shapes

z2014:
## Summary
- expose the OOXML `p:cNvSpPr/@txBox` marker as `ShapeElement.isTextBox`
- omit false values so existing serialized shapes remain unchanged
- document that regular shapes may still contain text through `textBody`

## Specification
ECMA-376 section 19.3.1.21 defines `txBox` as the marker for a specialized text-box shape. Its absence does not mean the shape cannot contain text.

## Tests
- `cargo test -p pptx-parser` (262 passed)
- `cargo clippy -p pptx-parser --all-targets -- -D warnings`
- PPTX TypeScript project and declaration checks
- `cargo fmt --check`
- `git diff --check`


z2014:
@yukiyokotani cc

z2014:
@yukiyokotani cc

PR #1358, opened 2026-08-23
feat(docx): Add track changes hook and comments

dbuxton:
# Word tracked changes + comment balloons for the DOCX viewer

Hi! Long-time admirer of this project. This is everything on my fork's `main` that isn't yours yet: two DOCX features that go together, tracked changes (§17.13.5) and comments (§17.13.4). Happy to split it in two or rework any of it — just say the word.

## Why

Deleted text rendered as ordinary body text, so a deleted paragraph just sat there looking like part of the document — neither of Word's views, not Final and not All Markup. Moves were worse: `w:moveFrom`/`w:moveTo` weren't parsed at all, so both ends of a tracked move vanished from the model. And comments were data-only: `doc.comments` had the text, but the anchors (`commentRangeStart`/`End`, `commentReference`) fell through the parser's default arm, so there was no way to know where a comment pointed.

## Tracked changes

New `showTrackedChanges` axis on `LayoutOptions`. Off (the default) is **Final view**: deletions and moved-away text produce no segment at all, so line breaking and pagination see the accepted document. That's the one behaviour change to existing output, and I think it's the right default — it's what Word shows when you open the file. On is **All Markup**: author-coloured underline for insertions/moveTo, strikethrough for deletions/moveFrom, and a hairline change bar in the left margin beside every changed line. Author→colour is first-appearance order over the main story, computed once per markup session; the default variant never builds it.

Since hiding deletions changes line breaking, the flag is a genuine geometry-selecting input: it's in `layoutOptionsKey`, each value is its own cached variant, and it's threaded through render / collect / hit-test / find so everything picks the same one. A document with no revisions lays out byte-identically in both (test pins this). Worker mode rides the existing options spread, no new messages. Side effect: it activates the `word-track-change-decoration` and `word-track-change-author-palette` compatibility facts, which were dead code.

## Comments

Anchors parse as zero-effect paragraph marks (`DocParagraph.commentMarks`) — no runs produced, run splitting/coalescing untouched, so layout geometry is identical with or without comments and a comment-less paragraph still serialises byte-for-byte the same. `commentsExtended.xml` supplies reply threading and resolved state.

The opt-in `showComments` option draws Word-style margin balloons: ranges tinted on the page, a connector stub per anchor, replies nested in the parent balloon, resolved threads hidden, balloons stacked in anchor order and capped at ten lines with truncation toward the next one. Click a balloon or its tinted range to select the thread — it expands to a page-full of lines and scrolls beyond that. Off means genuinely off: zero comment DOM, no gutter, render output unchanged. Geometry lives in a pure module (`comment-margin-layout.ts`), same pattern as xlsx's `comment-popup.ts`.

## API

All additive except the Final-view default: `showTrackedChanges` on the render / collect / hit-test surfaces; `showComments` + `commentsGutterWidth` and runtime setters on both viewers; `DocxDocument.commentAnchorRanges()`; `DocComment.parentId` / `.resolved` / `.paragraphs`; baseline regenerated. Worth flagging: this effectively revives the `showTrackChanges` option you removed in 0.77 as non-functional, under a different name and actually wired up this time. `doc.revisions` keeps its ins/del-only shape, and the markdown projection drops both ends of a move (it did before too, since they were unparsed — output unchanged).

## Testing

~3000 tests pass under `pnpm vitest run packages/docx`; the conformance suites importing locally-unbuilt WASM fail in a fresh checkout, which is pre-existing. New suites cover both layout variants, markup decorations and change bars, the balloon layout model, the options and their runtime setters, and both viewers in main and worker mode. `conformance/generate.ts` gains two deterministic synthetic fixtures — stored zip, fixed authors and dates, no real-document content — covering every §17.13.5 revision kind and a §17.13.4 thread with a reply plus a resolved thread. New Storybook story too, "Tracked changes & comments", with live toggles. ast-grep clean, rustfmt applied, README matrix and CHANGELOG updated.

Commits split parser → layout axis → decorations → comment geometry → DOM → fixtures/docs, so commit-by-commit will read better than the squashed diff.


yukiyokotani:
Thank you for the substantial work on tracked changes and comments.
While integrating this PR for v0.81.0, I decided that the library should expose review data and geometry but should not own a particular comment UI. Applications may need document overlays, a side pane, an independent list, or a combination of these.
I’m sorry that I made this boundary decision unilaterally, without discussing it with you first. As part of the integration, I am removing the built-in balloons, gutter, and Viewer-level display controls.
The parsing, metadata, anchors, final-state rendering, and geometry from your contribution will remain. I am also adding an official guide for building application-owned review UIs from these APIs. Thank you for moving this area forward.

dbuxton:
Do you want me to make those changes? Sounds smart and I'm happy to pick up

dbuxton:
I guess the only pushback against this is that forcing comments into the wrapper UI means that the wrapper needs to somehow (in order to get parity with Word): 

- draw the comment bubble and gutter and so on (this is easy)
- draw the highlight on the text indicating what text the comment relates to
- draw some sort of line or connector between the bubble and the highlighted text

This seems like quite a lot of work and complexity (particularly drawing the highlights on the text). I can definitely see a case for some specialized review workflows that want to do this but I wonder if it makes sense to bundle the experience in this PR as a default which can be switched off, potentially with some injectable styles or similar.

Maybe I misunderstand the architecture here and I'm sure it's nothing Claude can't solve but if the goal of the project is essentially Word-parity then I would challenge where this boundary ends up?

That said, it's your project so happy to work out how to make whatever decision you make work!

yukiyokotani:
Thanks for raising this. After implementing an application-owned example, I agreed that asking every consumer to recreate this UI would be too complex. ScrollViewer should provide a default read-only comment UI while still allowing applications to replace it. I’m limiting the default UI to ScrollViewer because it owns the scrolling viewport and can reserve stable space for comment cards without imposing layout decisions on the lower-level viewer.
We’ve narrowed v0.81.0 to ChartEx and related fixes. This PR will move to the following minor release because the public API and responsibility boundaries need more time to settle. No changes are needed from you for now; I’ll continue the integration and preserve the attribution.

dbuxton:
Great! I have another change that is partly blocked by this (that improves time-to-first-paint for large docx, which is particularly difficult with tracked changes - I was seeing complex 30 page docs take 30 secs to render even with all flags switched on) which I'll wait to submit until this lands because otherwise it'll be a lot of churn. 

You can see the working copy here in my fork if you're interested: https://github.com/dbuxton/office-open-xml-viewer/pull/3

PR #1373, opened 2026-08-25
fix(docx): trim trailing first-fragment partial rows in table pagination

wadimk:
## Problem

Paginating certain DOCX documents containing long multi-page tables with vertically merged cells (`w:vMerge` row spans) fails with:

```
Error: Table footnote admission did not converge
```

thrown from the retry loop in `body-paginator.ts` — even for documents that contain **no footnotes at all**. The whole document fails to render.

Affected document class: a table whose rows are admitted across a page boundary after a completed partial (split) row, where a trailing row is admitted as a *partial* row against the leftover height budget, and that row participates in a `vMerge` span that gets truncated by the fragment boundary.

## Root cause

In `takeTableFragment` (`packages/docx/src/layout/table-pagination.ts`):

1. After a split row completes, subsequent rows are admitted by retained track height; the last admitted row may be a **partial** row selected against the remaining budget.
2. `materializeFragment` re-runs `layoutTable` on the selected rows. When a fragment truncates a `vMerge` span, `resolveRowHeights` (`table.ts`) relocates the span owner's height deficit into the fragment's **last span row** — so the trailing partial row can grow well past the budget its line selection was made against (observed: fragment materializes to ~769pt against a ~728pt fresh page).
3. The over-admission trim loop only popped trailing rows whose ranges were **all `whole`**, so it stopped immediately at the trailing partial row. The fresh-page rescue requires `available < freshPage`, which does not hold on an already-fresh page. The function therefore returned an over-page fragment with `requiresFreshPage` unset.
4. The body paginator can never admit such a fragment: its footnote-admission retry loop re-requests the identical fragment (`reservePt` is 0 with no footnotes, so the requested extent never changes), the fingerprint repeats, and the convergence guard throws.

## Fix

Per ECMA-376 §17.4.6 a table row may split across pages, but a row group whose merged-cell content cannot physically fit the remaining band must move to the next page. A trailing partial row that is still in its **first** fragment (`fragmentIndex === 0`) has emitted nothing on earlier pages, so deferring it whole is a legal trim boundary: no content is lost or duplicated and the cursor never moves backward.

The trim condition changes from "trailing all-whole source row" to "trailing first-fragment source row":

```ts
const trimmableSourceRow = last?.ownership === 'source'
  && last.fragmentIndex === 0;
if (!trimmableSourceRow || sourceCount <= 1) break;
```

After popping the partial row, the loop keeps trimming whole rows until the fragment fits; the deferred row group re-paginates from the next page. The existing `fragmentIndex === 0` and `sourceCount <= 1` guards keep protecting already-emitted content and forward progress, so the `'Table pagination cannot advance from a fresh page'` invariant is preserved, and the paginator's convergence guard remains as a safety net.

## Tests

- New unit test in `packages/docx/src/layout/table-pagination.test.ts` (`retained table pagination > trims a trailing partial row when a vMerge owner deficit relocates into the truncated span`) builds a minimal 4-row table (plain split row + 3-row `vMerge` span) that reproduces the exact mechanism: admitted 230pt into a 200pt fresh page with an unpoppable trailing partial row. It fails before the fix (`expected 230 to be <= 200`) and passes after; a second `take` then paginates the remainder to completion.
- Full `packages/docx/src` vitest suite: **3104 tests passed**.
- `tsc --build`: clean.
- Verified end-to-end by paginating a real-world multi-page document whose 48-row vertically merged table previously triggered the non-convergence throw; it now paginates to completion.

### DIFF
--- a/packages/docx/src/layout/table-pagination.test.ts
+++ b/packages/docx/src/layout/table-pagination.test.ts
@@ -10,6 +10,7 @@ import type {
   AcquiredParagraphLayoutInput,
   LayoutServices,
   ParagraphLayout,
+  TableCellLayoutInput,
   TableEdgeInputs,
   TableLayoutInput,
   TableRowLayoutInput,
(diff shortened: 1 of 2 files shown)


yukiyokotani (approved):
Adversarial review completed against the current main branch. The change is limited to retained DOCX table pagination and its focused regression test; no dependency, build-script, binary, network, dynamic-execution, environment, or credential-handling changes are present. The first-fragment/source-row guard prevents rewinding already-emitted content, while the source-count guard preserves forward progress. Verified after a clean merge with current main: 3,163 DOCX tests, focused pagination tests, typecheck, package-build isolation, layout-boundary, compatibility-evidence, and public-API gates all pass. APPROVE.

PR #1374, opened 2026-08-25
feat(docx): paint insertions (w:ins) with author-coloured underline

wadimk:
## Problem

The DOCX model carries per-run revision facts (ECMA-376 §17.13.5 `<w:ins>` / `<w:del>` — `DocxTextRun.revision`) end-to-end, but the paint pipeline never converts them into `TextDecorationLayout` entries. Tracked insertions render indistinguishable from accepted content, and both the README feature table and the `RunRevision` doc comment claim author-coloured inline markup that was never wired.

## What this does

Builds revision decorations where other run decorations are built, matching the insertion presentation of Word's **All Markup** view:

- **Insertions (`<w:ins>`)** gain an author-coloured underline.
- Revision decorations are appended **after** authored ones, so a run that is both underlined (§17.3.2.40 `<w:u>`) and inserted shows both rules, with the author color visible.
- Decoration style/thickness reuse the same selected-face probes as authored `<w:u>`; the run's own text color is untouched.

**Author color policy.** Word's author-color assignment is renderer-defined (per-installation, not document-authored), so this uses a deterministic FNV-1a hash of the author name into an eight-color palette (Word's author-color set) — stable across pages, sessions, and measure order. Absent/empty author takes slot 0. More than 8 distinct authors repeat slots, the same class of behavior as Word.

The palette, the kind→decoration mapping, and the author-color hash live in a new production module `layout/track-changes.ts` (the earlier test-only scaffolding in `paint-compatibility.ts` was removed in #1372; this is its first production consumer).

## Interaction with #1372 (accepted-final-state projection)

Since #1372, `<w:del>` / `<w:moveFrom>` runs produce no canvas segments at all — the canvas renders the accepted document, and revision metadata is reserved for consumer-owned review UIs. This PR builds on that decision:

- **Insertions** remain visible content and now get their author-coloured underline.
- **Deletions** stay projected out of the canvas (served to review UIs through the parsed revision metadata, §17.13.5).
- The deletion→strikethrough arm of `wordTrackChangeDecoration` / `retainedTextDecorations` is kept (symmetric with the policy mapping, unit-tested, reusable by a consumer review UI) but is unreachable in production while the projection stands — happy to trim it if you prefer.

## Tests

- New `layout/track-changes.test.ts` — palette pins, kind→decoration mapping, author-color determinism/membership/absent-author/distinct authors.
- `layout/retained-typography.test.ts` — insertion→underline in author color with exact geometry; authored underline + insertion yields both decorations.
- `track-changes-render.test.ts` (recording-canvas via `renderDocumentToCanvas`) — insertion strokes in author colors, no decoration without revision, per-author colors, stacking on authored underline, text fill untouched; deleted text neither paints nor carries decorations (projection pinned).

## Verification

- `npx vitest run packages/docx/src` — all green (incl. the #1372 conformance suite; requires a local `pnpm build:wasm` after the rebase since the parser changed).
- `tsc --build` clean; `ast-grep scan` clean.
- Headless renders (node session API + software canvas) of a synthetic tracked-changes document and a real-world document with tracked insertions/deletions: insertions show author-coloured underlines in the right places, deletions follow the accepted-final-state projection, body text otherwise unchanged.

Co-Authored-By: Kimi <noreply@moonshot.ai>

### DIFF
--- a/packages/docx/src/layout/paragraph.ts
+++ b/packages/docx/src/layout/paragraph.ts
@@ -73,6 +73,10 @@ import {
   type RetainedEmphasisClusterInk,
   type RetainedEmphasisMarkInput,
 } from './retained-typography.js';
+import {
+  wordTrackChangeAuthorColor,
+  wordTrackChangeDecoration,
+} from './track-changes.js';
 import type { RunTypographyAcquisitionInput } from './typography-input.js';
 import { resolveAnchorFrame, type AnchorReferenceFramesInput, type AnchorFrameResult } from './anchor-frame.js';
 import { paragraphGapPt } from './paragraph-spacing.js';
@@ -208,6 +212,13 @@ interface RetainedTextGeometryPlan {
     double: boolean;
     probe: RetainedInkMetric;
     doubleProbe?: RetainedInkMetric;
+  }>;
+  /** ECMA-376 §17.13.5 tracked-changes markup in the deterministic
+   * revision-author color, painted in addition to authored decorations. */
+  readonly revision?: Readonly<{
+    color: string;
+    underline?: Readonly<{ probe: RetainedInkMetric }>;
+    strike?: Readonly<{ probe: RetainedInkMetric }>;
   }>;
   readonly emphasis?: Readonly<{
     authored: string;
@@ -913,6 +924,7 @@ export function planLine(input: PlanLineInput): LineLayout {
             color: retainedColorString(style.color),
             ...(retainedGeometry.underline ? { underline: retainedGeometry.underline } : {}),
             ...(retainedGeometry.strike ? { strike: retainedGeometry.strike } : {}),
+            ...(retainedGeometry.revision ? { revision: retainedGeometry.revision } : {}),
           })
         : style.decorations;
       const emphasis = retainedGeometry?.emphasis ? {
@@ -1680,8 +1692,10 @@ function retainedGeometryPlan(
   sourceOffset: number,
   color: TextPlacement['color'],
 ): RetainedTextGeometryPlan | undefined {
+  const revisionDecoration = wordTrackChangeDecoration(segment.revision?.kind);
   if (!(segment.highlight || segment.underline || segment.strikethrough
-    || segment.doubleStrikethrough || segment.emphasisMark)) return undefined;
+    || segment.doubleStrikethrough || segment.emphasisMark
+    || revisionDecoration.underline || revisionDecoration.strike)) return undefined;
   const service = segment.textLayoutService;
   const request = segment.textShapeRequest;
   if (!service || !request) {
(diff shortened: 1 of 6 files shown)


yukiyokotani (changes requested):
Adversarial review: REQUEST_CHANGES. This implementation is based on the pre-#1377 rendering model and conflicts with current main, which deliberately projects tracked revisions as accepted-final: insertion/moveTo content is retained and deletion/moveFrom content is removed before layout. Reintroducing deletion glyphs unconditionally would reverse that default without an explicit markup-view API. The proposed FNV author-name hash is also not Word-compatible: Microsoft documents that By author colors are application/user settings and may change after reopening or on another computer; insertion/deletion formats are configurable, and inline versus balloon presentation depends on the active markup mode. Word defaults for moves also differ (double strike/double underline), which this change does not implement. Please redesign this as an opt-in markup projection with configurable display policy while preserving accepted-final as the default, then rebase on current main and add tests for final/markup modes, deletions, moves, inline/balloon policy boundaries, and author-color configuration. Security review found no added dependencies, scripts, binaries, external I/O, dynamic execution, environment access, or credential handling in this diff.

yukiyokotani:
## Evidence supporting the requested changes

- **Word does not use one fixed appearance for tracked changes.**
  - The colors and formatting used for insertions, deletions, and change bars are configurable.
  - Author colors are not stable identifiers; they may change when the document is opened on another computer.
  - Reference: [Microsoft Support — Change the track changes color](https://support.microsoft.com/en-US/Word/change-the-track-changes-color)

- **Word provides multiple review display modes.**
  - Simple Markup, All Markup, No Markup, and Original do not all render revisions directly in the document body.
  - Reference: [Microsoft Support — Track changes in Word](https://support.microsoft.com/en-us/word/training/track-changes-in-word)

- **The current renderer intentionally uses an accepted-final projection.**
  - Insertions and `moveTo` content are retained, while deletions and `moveFrom` content are removed before layout.
  - Regression test: [`tracked-changes-final-layout.test.ts`](https://github.com/yukiyokotani/office-open-xml-viewer/blob/65af367a47434e5b1aba004cde59381aa4e81da7/packages/docx/src/tracked-changes-final-layout.test.ts#L92-L124)
  - Implementation: [`line-layout.ts`](https://github.com/yukiyokotani/office-open-xml-viewer/blob/65af367a47434e5b1aba004cde59381aa4e81da7/packages/docx/src/line-layout.ts#L3278-L3290)

- **Therefore, the current approach in this PR cannot be accepted as-is.**
  - Always painting tracked changes with colors, underlines, and strikethroughs would implicitly replace the existing accepted-final behavior with something resembling All Markup.
  - Generating a deterministic color from an author-name hash is also not supported by Word's documented behavior.

### Requested revisions

- Keep accepted-final as the default behavior.
- If revision markup is added, expose it as an explicit display mode.
- Define colors, decorations, `moveFrom` / `moveTo`, and balloon / inline behavior from documented or verified Word behavior.
- Add tests for each supported display mode.

### Maintainer follow-up option

- If you prefer, the maintainers can take over the remaining design and implementation work in a separate follow-up.
- Please let us know whether you would like to revise this PR or hand the work over to us.
- Until that direction is confirmed, we will keep this PR on hold.

wadimk:
Thanks for the detailed review and the documented Word behavior references — that's a fair call.

We'll close this PR: the accepted-final projection is a reasonable renderer default, and our immediate need (showing deletions to users) is covered application-side via the revision metadata exposed by the engine (a consumer-owned review panel), which is exactly the integration point the projection was designed for.

Feel free to use any part of this branch (layout/track-changes.ts, the decoration plumbing in retained-typography.ts, the render tests) if it's useful for a future opt-in display-mode implementation — otherwise no worries. Thanks again for the thorough feedback.

PR #1384, opened 2026-08-26
perf(docx): Publish pages progressively rather than waiting for the whole document to be ready

dbuxton:
Layout is currently one blocking call: nothing appears until every page has been paginated, so a long document shows an empty viewer for as long as its whole layout takes. 

This PR makes pagination suspendable and uses that to paint the opening pages almost immediately, then grow the document in the background the way Word's page count settles during repagination.

Every pass in `body-paginator.ts` becomes a generator that yields between body entries — the one point where the pass holds no half-applied state — and `yield*` delegation carries that through the convergence solvers, so each convergence policy still exists once and the resumable and blocking paths cannot drift apart. On top of that, `progressiveLayout` lays out the first body entries as a genuine short document through the ordinary pagination path and publishes a growing prefix (each step 4x the last, capped at half the body) until the authoritative pass lands.

Time to first paint becomes roughly constant regardless of length: 366ms vs 7807ms on a 75-page document, 542ms vs 3098ms on 25 pages, 509ms vs 5084ms table-heavy. End-to-end cost of the extra prefix passes is 1.04x on 75 pages and 1.17x on 25; short documents skip intermediate steps entirely. Both `sliceLayout` and `progressiveLayout` are opt-in and off by default, so existing callers see no change in scheduling. 

Several pre-existing bugs that this makes load-bearing are fixed alongside: load built the DEFAULT layout variant while the viewer rendered the markup variant, so a reviewed document repaginated synchronously on first render (~11s vs ~2s); `_getLayout` returned geometry for a variant nobody was looking at, blanking pages when the markup view was shorter; `onVisiblePageChange` was latched on `topIndex` alone, so a "page X of Y" indicator stayed stranded on the preview count; `destroy()` did not abort background layout; and a superseded `load()` finishing in the background invalidated the document that had replaced it.

Public surface: `onVisiblePageChange` gains a third argument, `layoutComplete`, so a consumer can tell a settled total from a provisional one — backward-compatible for existing two-arg consumers. `layoutComplete` / `whenLayoutComplete()` are exposed on both `DocxDocument` and `DocxScrollViewer`, and `setDocumentLayoutValidation` lets an embedder re-enable retained-layout contract checks against a production build. README, site API reference and the public API baseline are updated in step.

Verification: an adversarial scheduler that suspends at every opportunity produces a `layoutFingerprint` identical to the blocking layout across all five synthetic shapes, and every published prefix page is compared page-by-page against the finished layout. Full docx suite passing (3131), `tsc --build` clean, `pnpm bench:layout` unchanged. The viewer half would benefit from a VRT run against your private corpus before release.

### Known bugs/limitations

The only thing I haven't been able to do is make page numbering on initial render reflect the real length of the document - it now starts with e.g. `Page 1 of 2` while the rest of the document loads and only updates in the background. 

However my recollection is that Word itself does something similar so I think this is acceptable.

Progressive layout applies to `mode: 'main'`; worker mode is unchanged. It could be moved to worker mode but it's more complicated and less beneficial. Let me know if you want me to look at that, @yukiyokotani


yukiyokotani (changes requested):
I independently re-reviewed the current head (`63dfaece`) from a clean detached worktree instead of relying on the earlier notes. The progressive-layout direction is promising, and the existing DOCX suite is green, but I can still reproduce four blockers.

1. **[P1] A prefix published as `exact: true` can differ from the authoritative layout.**

   `progressive.ts` assumes that only the final page of a truncated layout can change and drops exactly that page. That does not hold for `keepNext`: the paginator scans an arbitrarily long successor chain and can move the chain's first block from an earlier page when a successor beyond the truncation cut becomes available.

   Deterministic repro: 80 synthetic plain paragraphs, 20 words each, with paragraphs 44–47 marked `keepNext`. The first preview covers 48 entries and publishes two pages with `exact: true`, but the fingerprint of published page index 1 differs from final page index 1. This uses no private sample. The relevant paths are [`progressive.ts` lines 20–33 and 133–159](https://github.com/yukiyokotani/office-open-xml-viewer/blob/63dfaece3c2a16290f842db6f95a9f02d7a5999b/packages/docx/src/layout/progressive.ts#L20-L33) and the successor scan in [`body-paginator.ts` lines 1186–1239](https://github.com/yukiyokotani/office-open-xml-viewer/blob/63dfaece3c2a16290f842db6f95a9f02d7a5999b/packages/docx/src/layout/body-paginator.ts#L1186-L1239).

   At minimum, prefix publications need to remain provisional/inexact and repaint on replacement unless the paginator exposes a genuinely stable checkpoint. Please add this case as a regression test.

2. **[P1] A failed reload permanently suppresses completion of the retained progressive document.**

   `DocxScrollViewer.load()` increments `_loadGeneration` before acquiring the replacement, while `TerminalResourceOwner.replace()` intentionally retains the current document if that acquisition rejects. The retained document's callbacks still capture the previous generation, so [`onLayoutPartial` / `onLayoutComplete`](https://github.com/yukiyokotani/office-open-xml-viewer/blob/63dfaece3c2a16290f842db6f95a9f02d7a5999b/packages/docx/src/scroll-viewer.ts#L594-L640) ignore all later publications.

   Deterministic repro: load a two-page progressive fake document, let a replacement load reject, grow the retained document to 80 pages, then invoke its captured completion callback. `viewer.pageCount` reads 80 from the engine, but the spacer height remains at the two-page value. The generation needs to track the successfully installed owner (or be restored safely on a failed replacement), with a regression covering this atomic-swap path.

3. **[P1] `LoadOptions.currentDate` does not select the same variant for direct `DocxDocument.renderPage()` calls.**

   Load primes and records the explicit variant at [`document.ts` lines 468–472](https://github.com/yukiyokotani/office-open-xml-viewer/blob/63dfaece3c2a16290f842db6f95a9f02d7a5999b/packages/docx/src/document.ts#L468-L472), and the option's public documentation says this prevents the first render from synchronously repaginating. However, [`renderPage()`](https://github.com/yukiyokotani/office-open-xml-viewer/blob/63dfaece3c2a16290f842db6f95a9f02d7a5999b/packages/docx/src/document.ts#L971-L992) forwards no active load-time date when the per-render option is omitted; the renderer consequently selects the default variant.

   I reproduced this by priming only an explicit-date variant and calling `renderPage(canvas, 0)` without another `currentDate`: the default variant is built synchronously before the paint promise returns. This defeats progressive loading for the direct document API and can make geometry access and paint select different variants. Omitted render options should select the document's active load-time variant.

4. **[P1] The validation optimization violates the final architecture contract and currently fails the required boundary gate.**

   With validation disabled, [`assertAndDeepFreezeDocumentLayout`](https://github.com/yukiyokotani/office-open-xml-viewer/blob/63dfaece3c2a16290f842db6f95a9f02d7a5999b/packages/docx/src/layout/invariants.ts#L1073-L1087) accepts a retained layout containing `NaN`; the PR's own test explicitly pins that behavior. The architecture contract requires non-finite geometry and invariant violations to remain fatal. Separately, the new runtime import from `plain-data.ts` crosses the occurrence-projection boundary. The required command fails exactly as follows:

   ```text
   $ mise exec -- node scripts/check-docx-layout-boundaries.mjs --final
   OCCURRENCE_PROJECTION_RUNTIME_DEPENDENCY: packages/docx/src/layout/plain-data.ts -> ./validation-policy.js
   ```

   Please retain fatal geometry/invariant validation and remove the forbidden dependency rather than weakening the final retained-layout boundary. Also, `validation-policy.ts` states that `DocxDocument.load({ debug: true })` enables the switch, but there is no production call that does so.

Independent verification on this head:

- focused PR tests: 52 passed;
- full `packages/docx/src` suite: 3,218 passed;
- typecheck, boundary unit tests, compatibility evidence, public API, package-build tests, and `git diff --check`: passed;
- final architecture boundary: failed with the diagnostic above;
- the required merge-base private self-VRT is still outstanding and should be run after the fixes, with every changed page adjudicated against Office output per the repository workflow.

If it would be smoother and avoid several review/fix round trips, I am also happy to take over the remaining work from this head: add the regression tests, implement the lifecycle/variant/validation fixes, run the full gates and private VRT, and finish the PR. Alternatively, please address the findings here and I will re-review. Either path is fine.

**REQUEST_CHANGES**

dbuxton:
I'm happy to do it as I have some good validation docx docs to QA against locally @yukiyokotani

dbuxton:
@yukiyokotani one or two other things I'd like to add before you merge here although the specific issues you identified are fixed.

Main one is that comments now all load on the first page on progressive paint which isn't very nice.

Will let you know when ready.

dbuxton:
I'm happy with this. Performance is still not amazing on huge docs but maybe not much that can be done there. This is much more useful though as you don't have to wait 20s to preview the first page... @yukiyokotani lmk if any questions or feedback

yukiyokotani (dismissed):
I re-reviewed the complete PR at `c23a8579b8048b4ebf8f470034d77b107f2f53d0` from a clean worktree, including the new worker-progressive and tracked-changes paths.

The four findings from my previous review are fixed: provisional prefixes are no longer claimed exact, retained-layout validation remains fatal with debug validation disabled, optionless render calls follow the active variant, and the simple failed-reload case resumes callbacks for the retained document. Thank you for addressing those.

I can still reproduce two correctness blockers:

1. **[P1] A failed newer reload can permanently detach the retained progressive document's callbacks.** [`load()` rolls `_loadGeneration` back by one when its acquisition rejects](https://github.com/yukiyokotani/office-open-xml-viewer/blob/c23a8579b8048b4ebf8f470034d77b107f2f53d0/packages/docx/src/scroll-viewer.ts#L601-L698), but [`TerminalResourceOwner` keeps its own replacement generation monotonic](https://github.com/yukiyokotani/office-open-xml-viewer/blob/c23a8579b8048b4ebf8f470034d77b107f2f53d0/packages/core/src/internal/canvas-viewer-mechanics.ts#L365-L385). In an adversarial three-load test: progressive document A is installed with generation 1; B starts; C starts and rejects; the catch restores the viewer generation from 3 to 2; B then resolves but the owner correctly discards it as stale because the owner's generation is still 3. A is therefore still installed, but its callbacks captured generation 1 and now reject its later partial/final publications because the viewer remains at 2. I reproduced A completing from a 2-page prefix to 80 pages while the scroll spacer stayed at the 2-page height. The request counter should remain monotonic; use a separate installed-document identity/token for callback admission (or otherwise restore the exact retained document's callback authority), and add this B/C overlap sequence as a regression test.

2. **[P1] Runtime tracked-changes switching in worker mode paints one variant while exposing geometry from another.** [`setLayoutView()` only changes `activeLayoutOptions`](https://github.com/yukiyokotani/office-open-xml-viewer/blob/c23a8579b8048b4ebf8f470034d77b107f2f53d0/packages/docx/src/document.ts#L1304-L1319); it does not replace the worker metadata. However, [`pageCount`](https://github.com/yukiyokotani/office-open-xml-viewer/blob/c23a8579b8048b4ebf8f470034d77b107f2f53d0/packages/docx/src/document.ts#L1116-L1120) and [`pageSize()`](https://github.com/yukiyokotani/office-open-xml-viewer/blob/c23a8579b8048b4ebf8f470034d77b107f2f53d0/packages/docx/src/document.ts#L1362-L1371) keep returning `_meta` from the load-time variant, while render requests are filled from the newly active variant. With a synthetic tracked document whose final layout is 11 pages and markup layout is 13, `setLayoutView({ showTrackedChanges: true })` leaves `pageCount === 11` while subsequent paint requests select the 13-page markup layout. The viewer then computes clamping, spacer height, and mount/recycle decisions from stale geometry. Worker-mode variant switching needs to obtain and atomically install metadata for the selected layout before viewers read geometry (an async switch is fine if necessary). Please cover this with a real worker-protocol/`DocxDocument` test; the current fake-viewer tests manually update the fake engine's page count and therefore mask the defect.

One small, independently proven diagnostic issue remains as well: **[P3] the worker-silence error always says `undefinedms`**, because [`_onParseWentSilent()` clears `_parseWatchdogMs` before interpolating it](https://github.com/yukiyokotani/office-open-xml-viewer/blob/c23a8579b8048b4ebf8f470034d77b107f2f53d0/packages/docx/src/document.ts#L837-L851). Preserve the value before clearing and assert the exact timeout in the test.

Verification on this head:

- all required DOCX architecture gates, compatibility/public-API/package-build checks, typecheck, and `git diff --check`: pass
- full DOCX suite: 3,261 passed, 1 skipped
- full repository Vitest run: 8,328 passed; the six failures were only sandbox-denied loopback listeners, and both affected files passed 12/12 when rerun with loopback access
- DOCX browser conformance: 66/66 across Chromium, Firefox, and WebKit
- merge-base-pinned private self-VRT: 48 documents / 405 pages, 0 pixel differences (`b6a22cc0ffcf5707597156ba741cd87881c76bd3` baseline versus this head, using separate ports and matching input hashes)
- GitHub CI: green

All verification run on this head is currently passing. The request for changes is based on the correctness blockers above, not a failing validation gate.

**REQUEST_CHANGES**

yukiyokotani:
Thank you for the thorough follow-up. I really appreciate how carefully you addressed each review point and added focused regression tests.

yukiyokotani:
v0.83.0 has been released.

The "Try yours" experience on the official website now enables both progressive mode and worker mode. Opening pages appear quickly while the rest of the document is processed in the background, resulting in a much smoother and more responsive experience—even for demanding documents.

I would also like to thank you again for the PR that set this work in motion. Your thoughtful implementation and sincere engagement throughout the review provided the foundation for the improvements that followed. Thank you for making such a meaningful contribution to the project.

PR #1391, opened 2026-08-26
perf(docx): memoize processed graphs in snapshotPlainData

wadimk:
## Problem

Every layout-service boundary in the DOCX engine snapshots its input via `snapshotPlainData` with a full deep **validate → `structuredClone` → deep freeze**. Because table acquisition/layout inputs embed previously snapshotted (already immutable) graphs, the same multi-megabyte structures are re-validated, re-cloned, and re-frozen at every boundary.

Measured on a numbering/styles-heavy document (11 pages, 137 paragraphs, 115 styles, 25 abstractNum / 225 numbering levels; parsed model itself is small — 0.5 MB JSON, parse time 66 ms, full pagination 90 ms):

- **`openDocxDocument`: ~18 s**, of which ≈75% is the snapshot trio (validate 5.3 s + `structuredClone` 5.2 s + freeze 3.3 s).
- Instrumented `structuredClone`: **~850 calls × ~15.8 MB ≈ 1.4 GB cloned** per open; dominant labels: `TableLayoutResult` (64×), `TableLayoutInput` (64×), `RetainedTableAcquisition.input` (24×).
- Node-walk counter: **99.5% of re-validated nodes were already frozen** — previously processed graphs being re-processed wholesale.

## Fix

Memoize processed (validated + deeply frozen) graphs in a module-level `WeakSet` in `layout/plain-data.ts`:

- A registered root passed to `snapshotPlainData` is returned as-is.
- Validate/clone/freeze are fused into a single walk that only touches genuinely new nodes; **registered subtrees inside fresh wrappers are reused by reference** (safe: they are immutable and engine-owned — caller-owned input nodes are never reused, so later caller mutation of inputs cannot leak into snapshots).
- `deepFreezePlainData` separately skips graphs it already froze (distinct freeze-only registry that **never relaxes validation**: frozen-but-never-validated graphs are still fully checked — pinned by a test).

Public contract unchanged: outputs deeply frozen, inputs never mutated, same `TypeError`s for non-plain/cyclic data, DAG identity preserved, own `__proto__` data properties handled with `defineProperty` (`structuredClone` semantics).

## Tests

- 6 new focused tests in `layout/plain-data.test.ts`: identical reference for an already-processed graph; fresh wrapper reuses a processed subtree by reference and is itself frozen; new invalid data next to a processed subtree still throws; frozen-but-unvalidated graphs still validated; sealed data counts as processed; `__proto__` edge case without prototype pollution.
- Full docx suite: **3169 tests green**; `ast-grep scan` clean. (`tsc --build` shows 16 errors in `packages/core` that are pre-existing on `origin/main` — identical count with the change stashed.)

## Measured impact (headless node session API, same harness)

| Document | Before | After | Speedup |
|---|---|---|---|
| Numbering/styles-heavy, 11 pp | 17.9 s | **8.0 s** | 2.25× |
| Table-heavy, 58 pp | 7.5 s | **5.2 s** | 1.4× |
| Small, 2 pp | 301 ms | 289 ms | ~1× (no regression) |

Page renders are **pixel-identical** (SHA-256 of rendered PNG) before/after.

## Scope note

The residual ~8 s is one validate+clone+freeze pass over genuinely fresh per-table layout graphs from `RetainedTableAcquisition` (24 acquisitions × ~162k fresh nodes, ~2% shared). Reducing that further needs acquisition-layer changes (caching retained acquisitions across pagination passes) — a separate, riskier PR.

Co-Authored-By: Kimi <noreply@moonshot.ai>


wadimk:
Closing as superseded by parallel upstream work.

Re-measured against current main (which now includes #1392 plus the fused clone+freeze traversal in 5d2ea63c and the development-only contract checks in bb9b0f44/8e34662b): on a numbering/styles-heavy 11-page document open time is 1.44s with main alone vs 1.46s with this PR's memoization applied on top — the delta is within noise. The development-only validation policy covers the same root cause (repeated validate/clone/freeze of immutable graphs at layout boundaries) more thoroughly than the WeakSet memoization proposed here.

Thanks for the thorough review culture on this repo — the profiling data in this PR (repeated ~15.8MB snapshots at every table layout boundary) may still be useful if you ever revisit the acquisition-layer residual costs.

yukiyokotani:
Maintainer follow-up is continuing in #1398 because GitHub rejected direct updates to this fork branch despite maintainer edits being enabled. #1398 preserves the original contributor commits and SHAs unchanged, merges current main normally, and adds separate maintainer fix commits. This PR will remain open until the replacement is fully verified and merged.

wadimk:
Follow-up noticed — thank you for carrying the work forward in #1398 (now merged) with the original commits preserved, and for the hardening fixes on top.

For the record: the earlier close was based on parity measurements against main at that moment (1.44s vs 1.46s on our reference document), not on any objection to the approach — glad the memoization landed in a verified form. We'll keep this PR closed since the replacement is merged.

Issue #1408, opened 2026-08-28
feat(tables): copy from docx & pptx tables similar to how copying from xlsx regions works


JanMarvin:
Hi @yukiyokotani ,

one feature that I would like to see, is being able to copy tables from docx and pptx similar to how i can copy cells from xlsx files. Right now tables behave slightly different in pptx and docx (unless I messed something up). The inputs were created with `rmarkdown`.

[showcase.docx](https://github.com/user-attachments/files/31565166/showcase.docx)

[created from this markdown file](https://github.com/JanMarvin/oxview/blob/main/inst/examples/showcase_docx.Rmd)
<img width="555" height="352" alt="Image" src="https://github.com/user-attachments/assets/b217903c-a2cb-4ee1-a3c7-0d12cee2c382" />
in `DocxScrollViewer`

[showcase.pptx](https://github.com/user-attachments/files/31565194/showcase.pptx)

 [created from this markdown file](https://github.com/JanMarvin/oxview/blob/main/inst/examples/showcase_pptx.Rmd)
<img width="646" height="359" alt="Image" src="https://github.com/user-attachments/assets/67a2805d-f2b8-4ed5-b230-1c739e275784" />
in `PptxScrollViewer`

yukiyokotani:
Thank you for the suggestion. When I first read this issue, I thought that allowing users to select and copy cell ranges from DOCX and PPTX tables in the same way as XLSX would be a useful feature.

However, after considering the selection behavior in more detail, I realized that XLSX and DOCX/PPTX have fundamentally different document structures. In XLSX, the entire selectable surface is a grid of cells, so the start and end points of a drag can always be represented as cell coordinates, producing an unambiguous rectangular range. In DOCX and PPTX, a table is only one type of content among ordinary text, shapes, and other elements, and a selection may begin outside the table.

One idea I considered was to treat a table as a single selectable unit when the selection starts outside it, while allowing XLSX-like range expansion when the selection starts inside a cell. However, this would remove the existing ability to start a text selection outside a table and end partway through it. There is also no way, without an explicit mode or UI affordance, to determine whether a drag beginning inside a cell is intended to select a range of cells or only part of the text within that cell.

The library is designed to remain largely headless, so I would also prefer not to introduce a built-in context menu, selection handles, or a dedicated table-selection mode to resolve this ambiguity. I could not find an intuitive and consistent behavior that would support structured cell selection without regressing the existing text-selection behavior.

[...]

PR #1411, opened 2026-08-30
fix(core): retain text selection across canvas gaps

yukiyokotani:
## Summary

- retain the last valid native text-selection endpoints while a DOCX or PPTX drag crosses blank canvas space
- share the behavior in core and wire both static and scroll viewers
- leave XLSX selection behavior unchanged
- keep document-level listeners active only for a primary drag that starts on a tagged text run

This is a follow-up to #1408. It does not add cell-range semantics or infer a nearest text position; it only preserves browser-produced run endpoints when the pointer leaves selectable overlay content.

## Verification

- core, DOCX, and PPTX Vitest suites: 7,298 passed, 3 skipped
- full TypeScript build
- lint (pre-existing warning-only tripwires remain)
- DOCX/PPTX drag-selection smoke tests in Chrome, Firefox, and WebKit


yukiyokotani:
Closing after manual UX validation. Programmatic restoration of the native Selection competed with the browser's ongoing drag recomputation over sparse overlays, causing unstable range flicker and jumps across unrelated paragraphs. The implementation has been fully reverted; XLSX was not in scope.

PR #1442, opened 2026-09-02
fix(xlsx): preserve worker draw module initialization

### LINKED ISSUE #1441: XLSX worker mode throws `Cannot read properties of undefined (reading 'thin')` on bordered sheets
## Summary

In worker mode, rendering any worksheet containing a bordered cell throws:

```
TypeError: Cannot read properties of undefined (reading 'thin')
```

The render promise rejects and the canvas keeps whatever was drawn previously, so
in a viewer with a sheet tab bar it looks like "switching sheets silently does
nothing". Main-thread mode is unaffected.

The failure only happens when **no optional renderer descriptors are supplied**.
Passing any optional renderer (e.g. `chartEx`) masks it.

## Affected versions

- Reproduced on `0.84.1` (released) and still present on `main`.
- Originally observed on `0.83.1`.

## Reproduction

Any workbook with at least one bordered cell:

```js
import { XlsxWorkbook } from '@silurus/ooxml/xlsx';

const wb = await XlsxWorkbook.load(bytes, { mode: 'worker' }); // no renderers
await wb.renderViewportToBitmap(
  0,
  { row: 0, col: 0, rows: 8, cols: 4 },
  { width: 320, height: 200, dpr: 1 },
); // rejects
```

**Expected:** the viewport renders.
**Actual:** rejects with the `TypeError` above.

Adding `{ mode: 'worker', chartEx }` makes the same call succeed, which is the
tell that this is an initialization-order problem rather than a style-parsing one.

## Root cause

`XLSX_BORDER_PX` in `packages/core/src/draw/dash.ts` is a module-level `const`
read by `xlsxBorderDashArray(style)`, which the XLSX renderer calls for every
border edge.

In the flattened production render worker, esbuild wraps that module in a lazy
`__esm` initializer but hoists the *reader functions* out of it, so the function
is callable before the table is assigned. Every call site of that initializer
sits inside another `__esm` initializer reachable only through the dynamic-import
shims for the optional renderers. `packages/xlsx/src/render-worker.ts` statically
imports `renderWorksheetViewport` and the `renderer.js` helpers, so with no
optional renderers the initializer never runs and the table stays `undefined`.

`thin` appears in the message because it is the border style Excel emits most and
it is deliberately absent from the dash table (it is a solid style), so it reaches
the lookup's miss path — V8 then names the key that was asked for.

The same stranded initializer also defers the shared pattern-fill cache, so
pattern fills are exposed to the same failure (`reading 'get'`).

This is the same class of defect as `e80e60dd` ("fix(pptx): preserve worker
renderer initialization"), which fixed `packages/pptx/src/render-worker.ts` only —
the XLSX worker still holds static imports.

## Why CI does not catch it

Every XLSX case in `tests/worker-dist/fixture.html` loads with
`{ mode: 'worker', ...renderers }`, and pulling in an optional renderer
initializes the shared draw module as a side effect. The public demo workbook
also has no borders.

## Impact

Worker-mode rendering of styled workbooks, which is the common case for
real-world spreadsheets. On a real-world workbook, 7 of 11 sheets failed to
render; the sheets that succeeded were the unbordered ones.

## Verification performed

Built from source and ran the production-worker Playwright suite both ways:

- without the fix: fixture reports `data-status="error"` with the message above
- with the fix: suite passes, and the previously failing workbook renders every
  sheet with pixel coverage matching the main-thread control

## Proposed fix

Load the renderer and its orchestrator through explicit module promises in
`packages/xlsx/src/render-worker.ts` so ESM initialization completes before the
first viewport render — mirroring `e80e60dd`. An awaited dynamic import makes
esbuild emit the initializer chain.

lukethacoder:
fixes #1441

The self-contained production XLSX worker flattens the renderer across optional renderer module boundaries. Its statically imported render entry points were hoisted past the initializer of the shared draw module, so the shared border dash table was still undefined when the first bordered cell was stroked. Rendering then threw
`Cannot read properties of undefined (reading 'thin')` and the canvas kept the previously rendered sheet.

`thin` surfaced in the message because it is the border style Excel emits most and it is deliberately absent from the dash table (it is solid), so it reaches the lookup's miss path. The same stranded initializer also defers the shared pattern-fill cache, so pattern fills were exposed to the same failure. Main-thread rendering was unaffected: there the table compiles to an eager top-level binding.

Load the renderer and its orchestrator through explicit module promises so ESM dependency initialization completes before the first viewport render. This mirrors the PPTX worker fix in e80e60dd; the XLSX worker still held static imports, so the hazard was untouched there.

Existing coverage could not catch this: every XLSX case in the production-worker fixtures supplies optional renderers, and pulling one in initializes the shared draw module as a side effect. Add a self-authored bordered workbook loaded with no optional renderer descriptors to the production-worker and Vite-consumer fixtures, and assert painted ink. Without the fix the load rejects and the fixture reports an error status before the ink assertion runs.

Verified against a private bordered workbook whose styled sheets failed to render in worker mode: 7 of 11 sheets threw before the change and all 11 render after it, with pixel coverage matching the main-thread control. The new bordered fixture reproduces the throw on the released bundle and renders cleanly once initialization is restored. Also verified with tsc --build (remaining errors are only the ungenerated local WASM bindings) and ast-grep scan.

### DIFF
--- a/packages/xlsx/src/render-worker.ts
+++ b/packages/xlsx/src/render-worker.ts
@@ -33,9 +33,7 @@ import {
   type PullSessionResponse,
   type WorkerSvgDecodeResponse,
 } from '@silurus/ooxml-core/worker';
-import { renderWorksheetViewport } from './render-orchestrator.js';
 import { workerRenderDeps } from './worker-render-deps.js';
-import { inheritSheetRenderCache, markAutoRowHeightsPrepared } from './renderer.js';
 import { XLSX_GOOGLE_FONTS, xlsxFontPreloadNames } from './google-fonts.js';
 import { resolveSharedStringRows } from './shared-strings.js';
 import {
@@ -82,6 +80,14 @@ const rawParts = new BoundedRawPartCache({
   maxEntries: HARD_MAX_RAW_PART_CACHE_ENTRIES,
   maxBytes: HARD_MAX_RAW_PART_CACHE_BYTES,
 });
+// Keep the renderer and its orchestrator behind explicit module boundaries. The
+// production worker is flattened into one self-contained asset, and a static
+// function import can otherwise be hoisted past the initializers of shared draw
+// dependencies that are also reached by optional renderers. Awaiting the modules
+// preserves ESM initialization order, so the shared border dash tables and
+// pattern-fill caches exist before the first bordered or filled cell is stroked.
+const rendererModule = import('./renderer.js');
+const orchestratorModule = import('./render-orchestrator.js');
 const worksheetPull = new WorksheetPullWorker(
   () => host.archive,
   (sheetIndex, worksheet, measured, resourceUsage) => {
@@ -237,6 +243,8 @@ self.onmessage = async (e: MessageEvent<
     if (req.type === 'renderViewport') {
       if (!workbook) throw new Error('Workbook not loaded');
       await fontsLoaded;
+      const { inheritSheetRenderCache, markAutoRowHeightsPrepared } = await rendererModule;
+      const { renderWorksheetViewport } = await orchestratorModule;
       const ws = sheetCache.get(req.sheetIndex);
       if (!ws) throw new Error('Worksheet is not loaded through its pull session');
       // Apply view-only size mutations to a render-local projection. Multiple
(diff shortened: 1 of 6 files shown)


yukiyokotani:
Thank you for the thorough fix! I really appreciate the detailed root-cause analysis and the regression coverage for both the published production worker and the Vite consumer rebundle.

I confirmed the reproduction from Issue #1441 and verified that the new tests fail without the fix and pass with it. Thanks again for your contribution!

PR #1444, opened 2026-09-02
feat(xlsx): preview delimited text in XlsxSheetViewer

danielrhodes:
## Summary

- add explicit CSV, TSV, and generic delimited-text preview loading to `XlsxSheetViewer`
- keep `XlsxWorkbook` and the composite `XlsxViewer` OOXML-only
- match XLSX URL / `ArrayBuffer`, reload, callbacks, worker mode, and `destroy()` ownership
- parse delimited text off the main thread and render it through the existing worksheet surface
- preserve every field as text and enforce bounded source and worksheet resource policies
- let Try Yours quietly select CSV and TSV and open them in the sheet-only surface, without advertising or accepting generic delimited-text extensions

## API and positioning

This is intentionally a secondary convenience attached to the Excel-style sheet viewer, not a general tabular-data model or a broadening of the library beyond its OOXML focus.

```ts
await sheet.load('/data.csv', { format: 'csv' });
await sheet.load('/data.tsv', { format: 'tsv' });
await sheet.load('/report.dat', {
  format: 'delimited-text',
  delimiter: '|',
});
```

The source contract is the same as XLSX loading: a string is fetched as a URL, and an `ArrayBuffer` remains caller-owned. A browser `File` can be passed via `await file.arrayBuffer()`. A successful reload replaces and disposes the previous source, and `destroy()` releases the active source.

`csv` and `tsv` provide comma and tab defaults; callers may override the delimiter for regional variants. `delimited-text` requires one explicit delimiter, so extensions such as `.txt`, `.dat`, and `.psv` do not imply a format. There is no extension or content sniffing and no filesystem-path interpretation.

Every field remains text. Leading zeroes, long identifiers, dates, and values beginning with `=` are displayed without type or formula inference.

Fixed-width PRN, DIF, and SYLK are intentionally out of scope because they have different grammars and should not be presented as delimiter variants.

## Implementation notes

- Main-thread rendering uses a short-lived parsing worker, terminated after the worksheet model is transferred.
- Worker rendering parses, retains, and renders in the existing XLSX render worker without initializing the OOXML parser WASM.
- URL bodies are streamed through a hard 64 MiB decoded-source ceiling; caller buffers are checked before copying, and Try Yours checks `File.size` before materialization.
- Input size, rows, logical cells, columns, retained UTF-8 content, serialized worksheet size, and cache usage are bounded.
- CR and CRLF inside quoted fields are normalized to the renderer's LF hard-break convention.
- Archive-only operations fail closed for a delimited-text-backed internal workbook.
- Try Yours routes only real `.csv` and `.tsv` extensions to `XlsxSheetViewer`; `.dat` and generic `delimited-text` remain library-API-only capabilities. Its OOXML-oriented `live · WASM` badge is hidden for those two secondary previews.

## Contributor provenance

Daniel Rhodes' original contribution commit remains in this PR's history. The implementation above replaces the public model-backed workbook API after maintainer review, while retaining the original contribution and authorship in the branch history.

## Verification

- focused parser, bounded acquisition, workbook lifecycle, sheet-viewer, and Try Yours tests: 86 passed
- full test suite: 8,898 passed, plus 2 private-corpus harness tests
- TypeScript project build, production package build, and production site build
- DOCX / XLSX / PPTX public API baselines and viewer API symmetry
- production worker bundle structure and browser startup: 2 passed
- XLSX main-vs-worker browser equivalence, including CSV: 3 passed with 0% XLSX pixel difference
- Try Yours browser file selection, sheet-only rendering, and badge behavior for CSV and TSV: 2 passed

## Adversarial review

An independent agent reviewed the complete diff for API boundaries, parser correctness, lifecycle cleanup, worker behavior, resource safety, bundling, and Try Yours integration. Its initial findings led to bounded streaming acquisition, caller/File preflights, quoted CR/CRLF normalization, compressed-response handling, and a shared dot-requiring site classifier. A second review found two boundary issues in those fixes; both were corrected. The final independent re-review found no actionable issues and assessed the reviewed scope as safe to merge.

Residual non-blocking constraints are documented by the implementation: bounded stream assembly temporarily retains chunks plus one contiguous buffer, Try Yours mirrors the library's 64 MiB source ceiling, and non-streaming Fetch polyfills are outside the ESNext/native-Fetch target.


yukiyokotani:
Thank you for putting this together. I had actually been thinking along very similar lines: reusing the existing cell and worksheet rendering machinery to enable lightweight previews of CSV, TSV, and other tabular data, much like standalone TIFF previewing is a small additional capability derived from functionality already needed for OOXML rendering.

I do have some thoughts about the public interface, particularly whether a sheet-level entry point would be a better fit than `XlsxWorkbook.fromModel`, and how we might leave room for worker-backed loading and rendering without positioning the project as a general-purpose delimited-text library. I may explore those ideas independently and follow up with a separate PR that takes a somewhat different interface approach.

Either way, I would like us to make this capability available soon. Thank you for moving the idea forward and demonstrating a viable path.

yukiyokotani:
Thanks for the contribution — this is very close to something I had already been considering.

After reviewing the public interface, I have updated this branch substantially around `XlsxSheetViewer.load()`. Delimited-text previews now follow the same source and lifecycle contract as XLSX loading, including URL / `ArrayBuffer` input, reload replacement, callbacks, worker mode, and teardown, while remaining clearly a secondary convenience of the OOXML viewer. CSV and TSV have explicit presets, and other delimiter-based text such as `.dat` can be loaded through `{ format: 'delimited-text', delimiter: ... }`.

I kept your original commit and authorship in the PR history even though the implementation itself was redesigned. I would like to include this capability in the next minor release.

PR #1460, opened 2026-09-02
feat(fonts): support font providers in main and worker modes

danielrhodes:
## Summary

- Add a `FontProvider` extension point for application-owned font sources.
- Support URL and `ArrayBuffer` font assets across DOCX, PPTX, and XLSX text and charts.
- Run providers on the main thread, including when rendering in a worker, then transfer owned font bytes through the worker protocol.
- Encapsulate the existing Google Fonts behavior in `GoogleFontsProvider` and route it through the same lifecycle.
- Keep `useGoogleFonts: true` as backward-compatible shorthand for the built-in provider.

## Motivation

The current Google Fonts integration requires the end user's browser to contact `fonts.googleapis.com` and `fonts.gstatic.com`. Those third-party requests disclose network metadata such as the user's IP address and User-Agent. Depending on the application's deployment, consent model, and legal basis, that may create GDPR and broader privacy-compliance concerns.

Consumers need to be able to serve fonts from private infrastructure instead: for example, an authenticated application endpoint, a same-origin asset service, or an in-memory source. That must also work in render-worker mode, where the worker may not have access to the application's authentication client or private network abstraction.

This change keeps the library offline by default and keeps its focus on OOXML rendering. Font acquisition remains an application policy supplied at load time.

## API

Applications can provide a resolver:

```ts
import { FontProvider, DocxDocument } from '@silurus/ooxml';

class PrivateFonts extends FontProvider {
  async resolve(families, { signal }) {
    return Promise.all(families.map(async family => ({
      family,
      source: {
        data: await fetch(`/api/fonts/${encodeURIComponent(family)}`, { signal })
          .then(response => response.arrayBuffer()),
      },
    })));
  }
}

const document = await DocxDocument.load(bytes, {
  mode: 'worker',
  fontProvider: new PrivateFonts(),
  fontFailure: 'error',
});
```

The provider always runs on the main thread. URL assets are fetched there, and owned bytes are transferred to the render worker before layout and paint. This supports authenticated/private sources without giving the worker credentials or coupling it to application networking.

`fontFailure` defaults to graceful local fallback and can be set to `error` for strict deployments. `fontProvider` and `useGoogleFonts` are mutually exclusive.

## Backward compatibility

- `useGoogleFonts: true` continues to enable the existing Office substitutions and script-aware Noto fallbacks.
- `GoogleFontsProvider` owns Google-specific CSS lookup, substitution, registration, sharing, and cleanup behind the generic provider contract.
- Existing low-level `preloadGoogleFonts` and `unloadGoogleFonts` exports remain available as deprecated compatibility APIs.
- Embedded OOXML fonts remain authoritative, installed authored fonts remain first in the fallback list, and application-provided faces use document-scoped aliases.
- Public declaration baselines are updated for the additive API; existing viewer signatures remain compatible.

## Main-thread and worker behavior

DOCX, PPTX, and XLSX now share one bounded provider session for resolution, byte ownership, `FontFace` registration, worker transfer, failure handling, and destruction. Progressive PPTX/DOCX loading and lazy XLSX worksheet discovery resolve newly encountered families incrementally.

Provider routes are applied to ordinary text and charts in both modes. DOCX chart routes are retained with the chart resource before paint, preserving the immutable layout-to-paint boundary. Registrations are deduplicated and reference-counted across documents and secondary viewer `FontFaceSet`s, and concurrent registration is serialized so cleanup cannot leak a retained face.

The existing hard limits on provider family count, face count, individual face size, total retained bytes, and font-operation duration continue to apply.

## Privacy

No third-party font request occurs unless `useGoogleFonts` or an explicitly network-backed provider is supplied. Private providers can keep all requests same-origin or return in-memory bytes, including in worker mode.


yukiyokotani:
Personal reminder: address #1497 as part of this PR before merging. Confirm that the built-in Google Fonts path supports a caller-configurable CSS origin, including redirect-aware relative font URLs, per-origin caching/isolation, DOCX/XLSX/PPTX, worker/progressive loading, and graceful fallback without retrying the public service. If the new FontProvider API supersedes the proposed option, document the equivalent configuration and retain focused regression coverage for the #1497 use cases.

danielrhodes:
@yukiyokotani for my use case, it's possible the customized CSS you introduced might solve the problem I was looking to solve for self hosting. Would this mean I would need a separate CSS file per hosted font? Or how do you imagine that working?

yukiyokotani:
@danielrhodes 
If by “customized CSS” you mean the CSS-origin work discussed in #1497, that hasn't actually been introduced as a public feature yet. I deferred that work because I think it should be considered together with the broader font-provider design in this PR, rather than adding Google Fonts-specific configuration to the shared load options.

At the moment, I think something along the lines of `FontProvider` is probably the better direction for self-hosting as well. Ideally, the viewer would tell the application which font faces it needs, and the application could decide how to resolve them — from a self-hosted URL, an authenticated/private service, in-memory data, etc. In that model, self-hosting would not necessarily require a separate CSS file per font, or even CSS at all.

That said, I'm not ready to commit to the current `FontProvider` API yet. Once this becomes a public API, we need to be careful about where the abstraction boundary sits: what information is passed to the provider, how multiple faces/weights/styles are represented, how loading and caching work, worker transfer, progressive loading, fallback behavior, cleanup, and so on. There are also many different font delivery models in the wider ecosystem, so I'd prefer not to accidentally bake assumptions from Google Fonts into a supposedly generic API.

So I think the general direction is right, but I want to spend more time on the interface before merging or releasing it. Because of that, I don't expect to have a general self-hosting solution immediately.

danielrhodes:
@yukiyokotani Ultimately the goal is to define the fallbacks. The reason I went with this font provider approach is because it was compatible with workers. However, I like the idea of just providing CSS which the worker can parse and understand too - it feels elegant and minimal. But maybe I am understanding your point to be: needing to create a Google fonts compatible CSS would not solve the dependency problem?

yukiyokotani:
@danielrhodes Self-hosting the CSS and font files would address the dependency on Google. My concern is about the scope of the interface, rather than whether CSS can solve that problem.

I agree that CSS could be a useful, simple option. One stylesheet can describe multiple families, weights, and styles, so there would be no need for a separate CSS file per font or a Google Fonts-compatible API.

Your fallback use case is important. I also want to support loading the actual font requested by a document when it isn't available locally, including from direct URLs, authenticated services, or in-memory data. I'm leaning toward a provider interface with CSS as one possible adapter, while the library handles loading, caching, registration, and worker communication consistently.

The design isn't settled yet. I want to avoid unnecessary font downloads and repeated round trips between threads, so working through the API and main/worker behavior will take some time. When it lands, I plan to introduce it as a breaking change with migration instructions rather than retain compatibility shortcuts.

PR #1465, opened 2026-09-03
fix(fonts): resolve native Noto CJK web font aliases

z2014:
## Summary

- Resolve Linux-native Noto CJK family names to the equivalent Google Fonts family during on-demand preloading.
- Use the same shared alias table in Canvas font stacks so measurement and painting select the registered web font.
- Cover SC, TC, JP, KR, and HK across PPTX, XLSX, and DOCX without changing OOXML document data.

## Why this mapping is required

Linux packages such as `fonts-noto-cjk` expose native family names such as `Noto Sans CJK SC`. Google Fonts serves the same regional Noto CJK variant under the shorter CSS family name `Noto Sans SC`.

CSS treats those names as different font families. Previously, the loader could download and register `Noto Sans SC`, while Canvas still requested the authored `Noto Sans CJK SC`. On browsers without the system font, Canvas therefore skipped the downloaded face and used a system fallback.

The authored family remains first in the Canvas stack. This preserves native font selection on hosts that have it. The Google Fonts family follows as an explicit browser fallback. Parser data and downloaded OOXML files keep their original font names.

## Hong Kong scope

`Noto Sans CJK HK` resolves to the official `Noto Sans HK` Google Fonts family through the same shared classification, preload, and Canvas-routing paths.

Google Fonts does not provide `Noto Serif HK`. The implementation therefore keeps `Noto Serif CJK HK` unchanged and does not infer a TC serif substitute.

## Call chain

1. PPTX, XLSX, and DOCX preflight collects authored font names and detected scripts.
2. The shared catalog resolves native Noto CJK aliases to existing Google Fonts CSS URLs.
3. The loader parses `@font-face`, registers WOFF2 faces in the active `FontFaceSet`, and waits before first render.
4. Canvas stacks place the authored family first and the registered Google family second.
5. Measurement and final paint use the same resolved stack.

## Behavior preserved

- Remote loading remains gated by `useGoogleFonts`.
- Fonts load only after an OOXML document requests them.
- Google Fonts CSS subsets and browser HTTP caching are unchanged.
- No fonts are installed into the operating system.
- Existing fallback and warning behavior remains active when loading fails.
- Existing Calibri/Carlito and Cambria/Caladea substitutions are unchanged.
- Non-HK CJK fallback chains do not gain the HK family.

## Verification

- Focused core, PPTX, XLSX, and DOCX font tests: 141 passed.
- TypeScript workspace build passed.
- `git diff --check` passed.

### DIFF
--- a/packages/core/src/fonts/preload.test.ts
+++ b/packages/core/src/fonts/preload.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect, afterEach, vi } from 'vitest';
 import { preloadGoogleFonts, unloadGoogleFonts, parseFontFaceRules, _resetCssCacheForTests, type FontPreloadEntry } from './preload.js';
 import { _resetFontRegistryForTests } from './font-registry.js';
+import { SCRIPT_GOOGLE_FONTS } from './scripts.js';
 
 const G = globalThis as Record<string, unknown>;
 const ORIG = { document: G.document, self: G.self, fetch: G.fetch, FontFace: G.FontFace };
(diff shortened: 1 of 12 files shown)


z2014:
Adds shared Noto CJK font aliases so browser rendering uses the correct Google Fonts families without changing the original OOXML font names. @yukiyokotani cc

yukiyokotani (changes requested):
Thank you for the contribution. The overall approach is sound: keeping the authored OOXML family first, resolving the Google Fonts family through a shared table, and using the same stack for measurement and painting is the right design.

Before merging, could you please cover or explicitly bound the Hong Kong variant? The official Noto CJK family set also includes HK, and Google Fonts provides `Noto Sans HK`. At present, `Noto Sans CJK HK` is absent from `GOOGLE_CJK_FONT_ALIASES` and `CJK_GOOGLE_FONTS`, while `classifyCjkFont("Noto Sans CJK HK")` returns `null`. As a result, the same preload and Canvas-routing failure remains for Hong Kong documents across DOCX, PPTX, and XLSX.

Please add `Noto Sans CJK HK` -> `Noto Sans HK` through the shared classification, preload, fallback-chain, and three-format test paths. Google Fonts does not appear to provide a corresponding `Noto Serif HK`, so please document that limitation rather than silently mapping the native serif HK face to TC without evidence.

I verified the current focused suite (128 tests), the workspace typecheck after rebuilding generated WASM, and `git diff --check`; all pass. Once the HK scope is addressed, we can complete the renderer self-VRT review.

z2014:
Addressed the HK scope in `7c83e9bf`.

- Added `Noto Sans CJK HK` → `Noto Sans HK` to the shared alias catalog.
- Added HK classification, on-demand preload, and Canvas fallback routing across DOCX, PPTX, and XLSX.
- Kept existing SC, TC, JP, and KR fallback chains unchanged, so non-HK documents do not request the HK family.
- Explicitly left `Noto Serif CJK HK` unmapped. Its HK serif fallback chain is empty rather than substituting `Noto Serif TC`.
- Added focused coverage for catalog lookup, trimming and case handling, FontFace registration, missing HK serif mapping, and all three renderer stacks.

Verification: 141 focused tests passed, the TypeScript workspace build passed, and `git diff --check` passed. The branch is rebased onto the latest `main` and the PR still contains one commit.

z2014:
@yukiyokotani Could you please review it again? Thank you!

yukiyokotani (changes requested):
Thank you for addressing the HK coverage and for explicitly keeping the unavailable HK serif web-font mapping out of the catalog. The Sans HK path now looks correct.

One blocking issue remains in the empty HK serif fallback path. `cjkFallbackChain('hk', 'serif')` intentionally returns `[]`, but PPTX still appends a separator for every non-null CJK result, and XLSX always appends a comma after `cjkPart`. The resulting family lists are invalid:

- PPTX: `"Noto Serif CJK HK", , "Noto Serif", "Noto Serif Hebrew", serif`
- XLSX: `"Noto Serif CJK HK", , "Calibri", ...`

Chrome rejects these Canvas `font` assignments and preserves the previously selected `ctx.font`, so HK serif runs can be measured and painted using the preceding run's font. DOCX does not have this problem because it assembles the fallback arrays before joining them.

Please make the separator conditional on the filtered chain being non-empty in both PPTX and XLSX, and add regression assertions that the HK serif result is a valid list (at minimum, no `/,\s*,/`). I reran the updated focused tests: all 136 existing tests pass, while focused adversarial tests for these two outputs fail as shown above. I also confirmed the assignment rejection in headless Chrome.

yukiyokotani:
Thank you again for the contribution and for promptly adding the HK alias coverage. While completing the final adversarial review, I found one remaining HK serif edge case: the intentionally empty HK serif fallback could leave an invalid double comma in the PPTX and XLSX Canvas font strings. I added focused regression tests and fixed the stack construction so empty fallback lists do not emit separators.

Because the available maintainer credentials could not push that follow-up commit to the fork branch, I preserved your original commit and authorship in #1476, added the follow-up as a separate maintainer commit, and ran the full test suite plus the DOCX/PPTX/XLSX private self-VRT against the exact merge base. All CI checks passed, and #1476 has now been merged with a merge commit.

Thank you for identifying and implementing the native Noto CJK alias support. I am closing this PR only because it has been superseded by the merged #1476.

PR #1471, opened 2026-09-03
fix(markdown): separate review comments from document text

yukiyokotani:
## Outcome

Markdown export remains a lightweight, best-effort full-text projection inside each format's existing parser WASM. It no longer attempts to reconstruct semantic groups or visual reading order from presentation geometry, and this PR no longer introduces separate Markdown packages or WASM binaries.

Review comments are now clearly separated from document content across DOCX, PPTX, and XLSX:

- collect comments once in a final `## Review comments` appendix instead of interleaving them with body text;
- render comment text as blockquotes and replies as nested blockquotes;
- retain explicit resolved/closed state;
- identify comments only with reliable authored context: DOCX comment ID, PPTX slide number and local ordinal, or XLSX sheet and cell reference;
- keep PPTX speaker notes with their slide and distinct from review comments;
- avoid geometry-based target inference or repair of malformed thread metadata.

The README and API documentation now set the intended expectation: this feature is useful for AI ingestion and search as full-text extraction, but its output is not an authoritative semantic, layout, or reading-order representation.

The design discussion and decision record are captured in [this PR comment](https://github.com/yukiyokotani/office-open-xml-viewer/pull/1471#issuecomment-5540653795).

## Verification

- complete DOCX, PPTX, and XLSX parser test suites
- `cargo test --workspace` (sandbox-restricted MCP socket cases rerun successfully with local socket access)
- `cargo clippy --workspace --all-targets -- -D warnings`
- TypeScript project build
- production library build with regenerated DOCX, PPTX, and XLSX WASM
- full JavaScript suite: 8,929 passed, 25 skipped
- private-corpus harness tests: 2 passed

## Adversarial review

- **Specification and inference:** no new OOXML interpretation or visual-layout heuristic is introduced. Thread relationships, status, slide identity, and cell references come from parsed metadata.
- **Malformed metadata:** DOCX orphaned or cyclic comment relationships remain visible at most once without inventing a parent; traversal cannot loop indefinitely.
- **Resource safety:** PPTX's existing Markdown byte ceiling is enforced against document text and the deferred comment appendix together. Other formats retain their existing archive and model limits.
- **Module boundaries:** comment formatting stays format-local in the existing parser crates; there are no standalone Markdown WASM wrappers or duplicated parse pipelines.
- **Cross-format wiring:** native, browser, worker, CLI, and MCP paths continue to use the same format parser projection for DOCX, PPTX, and XLSX.
- **Rendering regressions:** no Canvas parsing model used by rendering, layout, or paint changed, so renderer VRT is not applicable.

## Migration

No API migration is required. Existing `toMarkdown()` calls continue to work. Consumers that compare Markdown output should expect review comments to move into the final quoted appendix.


yukiyokotani:
## Design decision after review

The original goal of this feature is to give an AI the document's complete textual content as context. It is not to reconstruct an author's intended prose structure or to provide a canonical Office-to-Markdown conversion.

### What we learned

PresentationML frequently does not encode the semantic relationships that a human infers visually. Ungrouped text boxes may be associated by a background shape, relative placement, typography, or convention, but those signals are ambiguous. Geometry-based grouping and reading-order inference can improve selected layouts while confidently introducing false relationships in others. For AI context, a plausible but incorrect Markdown hierarchy is more harmful than an explicitly lossy text projection.

DOCX and XLSX contain more explicit logical structure, but Markdown can still overstate semantics: a table's first row is not necessarily a header, distant spreadsheet regions are not necessarily one table, and nested or drawing-hosted content may not have a faithful linear representation.

### Direction

- Keep conversion as a small, best-effort text projection bundled in each format's existing parser WASM.
- Do not publish separate Markdown WASM binaries or independent format entry points.
- Do not add geometry-based semantic grouping, inferred columns, inferred headings, inferred number/body associations, or document-specific compatibility rules.
- Preserve only reliable authored structure where available: document/slide/sheet boundaries, explicit title placeholders, paragraph order, explicit lists, tables, chart values, notes, and comments.
- Use stable source order when no explicit reading order exists. Ambiguity should remain visible instead of being hidden behind a confident inferred hierarchy.
- Describe the result as lossy, best-effort textual context rather than a faithful semantic conversion.

The measured cost of keeping the existing converter in the normal parser WASM is modest: approximately 60 KB gzip for PPTX, 10 KB for DOCX, and 19 KB for XLSX. By comparison, separate Markdown WASM binaries are approximately 580–630 KB gzip per format and cause an additional WASM instance and repeated archive parsing when an application uses both the Viewer and AI context extraction. The separate-WASM trade-off is therefore not worthwhile for this deliberately small companion capability.

### Review comments

Comments must never be interleaved with body text without a clear boundary. All formats should collect them in a final appendix:

```markdown
## Review comments

### Slide 3

> **Author**
>
> Comment text
>
>> **Reply author**
>>
>> Reply text

### Sheet1 — B4

> **Author**
>
> Cell comment
```

Location metadata should use only facts the file explicitly provides:

- PPTX: slide number, plus a target shape only when the relationship is explicit.
- XLSX: sheet name and cell reference.
- DOCX: comment id and anchored text/range.
- Replies: nested blockquotes.
- Resolved threads: an explicit `(resolved)` label.
- Speaker notes remain a separate `Speaker notes` section and are not formatted as review comments.

### Consequence for this PR

This PR should not be merged in its current form. Its semantic layout inference and standalone WASM/package isolation exceed the intended scope. The existing main-branch converter is closer to the desired architecture; any follow-up should be limited to simplifying misleading Markdown assumptions, improving text completeness, and standardizing the separated review-comment appendix across DOCX, XLSX, and PPTX.

PR #1493, opened 2026-09-06
fix(core,docx,pptx): round programmatic scroll targets forward

wadimk:
## Problem

Programmatic page and slide navigation can report the preceding item when a fractional item-start offset is quantized below its boundary by the scroll surface.

## Final fix

- Preserve exact visible-window boundaries for manual scrolling and arbitrary coordinate lookup.
- Round only programmatic item-start targets forward to a whole CSS pixel before assigning scrollTop.
- Keep the target clamped to maxTop, including fractional end-of-document positions.
- Share the target resolver in core and use it from the DOCX and PPTX scroll viewers.

The initial implementation added a global 1 CSS px epsilon to visible-window searches. Adversarial review rejected that scope because it changed ordinary manual-scroll attribution and PPTX coordinate-to-slide lookup. The final implementation confines quantization handling to scrollToPage and scrollToSlide.

## Tests and verification

- Focused core, DOCX, and PPTX scroll-viewer tests: 297 passed.
- Broad core, DOCX, and PPTX suites: 7,664 passed, 3 skipped.
- TypeScript build/typecheck: passed.
- Production build: passed.
- GitHub CI: audit, rust, smoke, and typecheck passed.

No Canvas painter, renderer, OOXML parser, or layout-output code changed, so image VRT is not applicable to this navigation-only behavior. Integer-quantizing scroll surfaces are covered directly by viewer integration tests.

## Security and adversarial review

- No reportable security findings or malware indicators.
- No dependency, lockfile, workflow, binary, executable-mode, symlink, submodule, encoded-payload, network/process-execution, or invisible-control-character changes.
- O(1), allocation-free target resolution; existing virtualization bounds remain unchanged.
- DOCX and PPTX integrations are covered; XLSX does not use this virtual-scroll path.

Spec: N/A — browser scroll-position compatibility, no OOXML behavior change.

### DIFF
--- a/packages/core/src/internal/virtual-scroll.ts
+++ b/packages/core/src/internal/virtual-scroll.ts
@@ -2,6 +2,7 @@ export {
   computeUniformVisibleWindow,
   computeVisibleWindow,
   createVirtualScrollGeometry,
+  resolveItemStartScrollTop,
   type VirtualScrollGeometry,
   type VisibleRange,
   type VisibleWindow,
--- a/packages/core/src/layout/virtual-scroll.test.ts
+++ b/packages/core/src/layout/virtual-scroll.test.ts
@@ -287,7 +287,8 @@ describe('fractional item-start navigation targets', () => {
   it('keeps a viewport just before a fractional boundary on the preceding item', () => {
     expect(computeVisibleRange(heights, 10, 110, 50, 0).topIndex).toBe(0);
     expect(computeVisibleRange(heights, 10, 220, 50, 0).topIndex).toBe(1);
-    expect(computeUniformVisibleWindow(3, 100.4, 10, 110, 50, 0).topIndex).toBe(0);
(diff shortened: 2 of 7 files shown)


wadimk:
Analysis of the failing typecheck job (two timeouts in `packages/node/src/docx-document-session.test.ts` — 5000ms limit):

- The change in this PR is confined to viewer-side scroll math (`packages/core/src/layout/virtual-scroll.ts` + viewer tests); the node session package never touches `computeVisibleWindow`.
- Locally on this branch: `vitest run packages/node/src/docx-document-session.test.ts` → 7/7 passed in 4 consecutive runs (one isolated flake on the very first run — the suite takes ~35s wall-clock and is timing-sensitive on loaded machines; the two tests that timed out in CI run ~1s each locally).
- The same tests pass on recent main CI runs.

This looks like pre-existing timing flakiness rather than a regression from this PR. Could you re-run the failed job when convenient? (I can't trigger it — reruns require admin rights.) Happy to dig further if it fails again consistently.

Issue #1499, opened 2026-09-08
feat(fonts): allow locale-aware CJK fallback selection


rwv:
When a document contains only shared Han characters and its font names do not identify a CJK region, the shared script preloader defaults to Japanese (`jp`). The embedding page and browser language preferences currently do not participate in that decision. On Chinese-language sites, this can select Japanese glyph variants for otherwise ambiguous text.

Add a shared load/viewer option:

```ts
cjkFallback?: 'auto' | 'sc' | 'tc' | 'hk' | 'jp' | 'kr';
```

An explicit region controls ambiguous fallback only; authored fonts and existing unambiguous CJK hints retain priority. `auto` (also the omitted default) should resolve the first usable CJK preference from `<html lang>`, then `navigator.languages`, then `navigator.language`, falling back to `jp` when none is available. Explicit script subtags should take precedence over region subtags; bare `zh` should use the documented product default `sc`.

Resolve browser preferences once per load and transfer the concrete result to workers. Font preloading and measurement/painting must use the same policy in DOCX, XLSX, and PPTX, with no process-global mutable preference. Non-browser callers can select an explicit region for reproducible output. This option must not enable Google Fonts network access implicitly.

This is an application font-substitution policy, not a new claim about normative OOXML language semantics. The default can change ambiguous Han glyphs on CJK-language hosts and should be documented.

Authored by GPT-6 Astra (OpenAI Codex).

PR #1517, opened 2026-09-15
fix(pptx): carry the resolved CJK fallback into main-mode parsing

### LINKED ISSUE #1515: fix(pptx): cjkFallback is dropped when a presentation loads in mode: 'main'
`cjkFallback` (#1499, #1500) reaches every PPTX load path except one: `mode: 'main'` without `progressiveLayout`. That path preloads Japanese faces regardless of the resolved region, while the renderer's canvas font stack asks for the resolved one.

### Where it breaks

The main-mode parse request has no field to carry the preference. `PptxWorkerRequest`'s `parse` variant (`packages/pptx/src/worker-protocol.ts:49-55`) declares `kind`, `id`, `buffer`, `resourcePolicy` and `progressiveLayout` only, so `packages/pptx/src/presentation.ts:478` cannot send it:

```ts
this._mode === 'worker'
  ? ({ kind: 'parse', id, buffer, resourcePolicy, useGoogleFonts, cjkFallback: this._cjkFallback, renderers } satisfies RenderWorkerRequest)
  : ({ kind: 'parse', id, buffer, resourcePolicy } satisfies PptxWorkerRequest),
```

The in-worker preflight is then built without the option (`packages/pptx/src/worker.ts:129-131`):

```ts
preflightBuilder = request.progressiveLayout
  ? null
  : new PresentationPreflightBuilder(bootstrap);
```

`PresentationPreflightBuilder` passes `options.cjkFallback` to `PptxFontPreloadAccumulator` (`packages/pptx/src/presentation-preflight.ts:549`), which hands it to `ScriptPreloadAccumulator`. With `undefined`, `preferredCjkLanguage()` returns `this.scriptCjkLanguage() ?? this.cjkLang ?? 'jp'` (`packages/core/src/fonts/scripts.ts:404`). A deck whose text is shared Han and whose theme fonts name no region therefore preloads `Noto Sans JP` / `Noto Serif JP`, whatever the host language or an explicit `cjkFallback: 'sc'` says.

`load()` consumes exactly that set: `packages/pptx/src/presentation.ts:430-437` preloads `pres._preflight.fontPreloadNames`.

The other three PPTX paths are correct, which is why the gap is easy to miss:

| mode | `progressiveLayout` | preflight built at | `cjkFallback` |
| --- | --- | --- | --- |
| `worker` | either | `packages/pptx/src/render-worker.ts:185` | applied |
| `main` | yes | `packages/pptx/src/presentation.ts:604` | applied |
| `main` | no | `packages/pptx/src/worker.ts:131` | **dropped** |

DOCX and XLSX are unaffected. Their main-mode loads return the parsed model to the window and compute preload names there with the resolved preference (`packages/docx/src/document.ts:557`, `packages/xlsx/src/workbook.ts:511`), so their `parse` requests never needed the field. PPTX alone derives the preload set inside the parse worker, so for PPTX the preference has to cross the protocol.

### Observable effect

On a `zh-CN` host loading an ambiguous-Han deck with `mode: 'main'`, `useGoogleFonts: true` and no explicit region, `resolveCjkFallback` yields `sc`, so painting asks for `Noto Sans SC, Noto Sans TC, Noto Sans JP, Noto Sans KR` (`packages/core/src/fonts/scripts.ts:232`, via `renderSlide`'s `cjkFallback: this._cjkFallback` at `packages/pptx/src/presentation.ts:1057`). The only registered faces are the JP ones, so the stack falls through to `Noto Sans JP` and Han renders with Japanese glyph variants — the exact outcome #1499 asked to avoid — after downloading a multi-megabyte family the stack did not want.

An explicit `cjkFallback: 'sc' | 'tc' | 'hk' | 'kr'` is ignored the same way, so the escape hatch documented for reproducible output does not work in this mode either.

### Proposed fix

1. Add `cjkFallback?: import('@silurus/ooxml-core').CjkLang` to the `parse` variant of `PptxWorkerRequest`.
2. Send `cjkFallback: this._cjkFallback` from the main-mode branch at `packages/pptx/src/presentation.ts:478`.
3. Build the preflight with it: `new PresentationPreflightBuilder(bootstrap, { cjkFallback: request.cjkFallback })`.

The resolved value is already a concrete `CjkLang` snapshot taken once in `load()` (`packages/pptx/src/presentation.ts:336`, `388`), so this transfers the same value the render worker already receives and introduces no second resolution site. Renderer, measurement and the Node API need no change.

### Acceptance

- A focused regression over the main-mode, non-progressive preflight: a Han-only deck loaded with `cjkFallback: 'sc'` preloads the SC faces rather than the JP ones.
- Worker mode, main + `progressiveLayout`, and `packages/node` keep their current behavior.
- `pnpm typecheck` and the PPTX test suite pass.

Authored by Claude Opus 5 (Claude Code).

rwv:
Fixes #1515.

`cjkFallback` (#1499, #1500) reached every PPTX load path except one: `mode: 'main'` without `progressiveLayout`, which is what an embedding application gets by default. There the resolved region never left Window, so the preflight that produces the Google-Fonts preload set fell back to `jp`.

### Root cause

`PptxWorkerRequest`'s `parse` variant had no field for the region, so `PptxPresentation._parse` could not send it and `worker.ts` built `new PresentationPreflightBuilder(bootstrap)` without it. `PptxFontPreloadAccumulator` then resolved ambiguous Han through `ScriptPreloadAccumulator.preferredCjkLanguage()` — `script evidence ?? cjkLang ?? 'jp'`.

`load()` consumes exactly that set (`preloadGoogleFonts(pres._preflight.fontPreloadNames, …)`), while `renderSlide` receives `this._cjkFallback`. On a `zh` host the canvas stack therefore asked for `Noto Sans SC, Noto Sans TC, Noto Sans JP, …` while the only registered faces were the JP pair: shared Han painted with Japanese glyph variants out of a family the stack did not want, after fetching it. An explicit `cjkFallback: 'sc' | 'tc' | 'hk' | 'kr'` was ignored in that mode too, so the documented escape hatch for reproducible output did not work either.

| mode | `progressiveLayout` | preflight built in | region |
| --- | --- | --- | --- |
| `worker` | either | `render-worker.ts` | applied |
| `main` | yes | Window, `_parseMainProgressively` | applied |
| `main` | no | `worker.ts` | **dropped** |

PPTX is the only affected format: DOCX and XLSX main-mode loads return the model to Window and compute preload names there (`docx/src/document.ts`, `xlsx/src/workbook.ts`), so their parse requests never needed the field. PPTX derives them from the preflight built inside the parse worker, so for PPTX the region has to cross the protocol in both modes.

### Fix

Add `cjkFallback` to the `parse` request, send the region `load()` already resolved once, and build the preflight with it.

The field is **required** rather than optional, which is what makes the original omission impossible to reintroduce: a parse request that drops the region no longer compiles. Both main-mode sites now carry it, including `_parseMainProgressively`, where the worker builds no preflight and the value is inert — this keeps the protocol's contents from depending on a worker-side implementation choice. `RenderWorkerRequest`, the Window side of progressive loading, and `packages/node` are untouched.

### Validation

- `./node_modules/.bin/vitest run packages/pptx/src/cjk-fallback-main-mode.test.ts` — the new regression drives `worker.ts` against a mocked archive through parse → open → pull → ack → finish and asserts the preload set. On the previous code it reports `[ 'Noto Sans JP', 'Noto Serif JP' ]`; with the fix, `[ 'Noto Sans SC', 'Noto Serif SC' ]`. Reverting only the `worker.ts` line reproduces the failure, and removing only the `presentation.ts` field is now a `TS1360` build error.
- `./node_modules/.bin/vitest run packages/pptx` — 121 files, 878 tests passed.
- `./node_modules/.bin/vitest run packages/pptx packages/core` — 270 files, 4295 tests passed.
- `./node_modules/.bin/tsc --build --pretty false` — clean.
- `npx ast-grep scan` — 0 errors; the 81 pre-existing `no-as-unknown-as-cast` tripwire warnings are unchanged and none are in the touched files.
- `git diff --check` — clean.
- Not run here: `pnpm build:wasm`, VRT, and the WASM-backed suites. This environment has no Rust toolchain, so the generated parsers under `packages/*/src/wasm/` are absent. The full `vitest run` fails 18 files / 19 tests for that reason (`packages/node`, `packages/docx/src/conformance`, `packages/docx/src/render-worker-layout-parity`, `packages/vscode-extension/src/mathEngineBundle`); re-running exactly those 18 files on the unmodified base gives the identical 18 failed / 19 failed / 13 passed, so they pre-date this branch. The change touches no parser, layout or painting code.

### Adversarial review

- **Specification fidelity.** No OOXML rule changes. Regional Han fallback is library font-substitution policy as established in #1499; this only makes main mode honour the policy the renderer already applies, so parse-worker and Window agree on one region.
- **No unexplained heuristics.** Nothing is inferred or thresholded: the value transferred is the same `resolveCjkFallback()` snapshot the render worker already receives, taken once in `load()`.
- **Cross-format wiring.** DOCX/XLSX inspected and deliberately excluded (their main-mode preload names are computed in Window from the returned model); `packages/node` resolves and applies its own region and does not use this protocol; viewer and scroll-viewer already forward `opts.cjkFallback` into `load()`. The format-level reason is recorded next to the new field in `worker-protocol.ts`.
- **Performance and resource safety.** One short string on an existing message; the transfer list is unchanged, and the options literal is not retained (`PresentationPreflightBuilder` copies the scalar out). For a deck whose text is uniformly ambiguous Han the preload set is a 1-for-1 swap of two Noto faces (`hk` is cheaper still, sans-only). A deck that mixes kana/hangul with ambiguous Han can now fetch one extra family, because `PptxFontPreloadAccumulator` unions per-slide script names by design — worker mode has always paid that, and main/worker preload parity is the point of the fix.
- **Behaviour change.** Main-mode decks with ambiguous Han on a CJK-language host now preload the resolved region's faces instead of the `jp` default. That is the outcome #1499 asked for, and it is already documented as the default's consequence.


yukiyokotani (approved):
Independent adversarial review found no blocking or non-blocking defects. The issue is valid: main-mode non-progressive PPTX preflight was built in the parse worker without the already-resolved regional CJK fallback, so ambiguous Han could preload JP faces while rendering requested another region. The patch carries the concrete value across the internal protocol, makes omission a type error, and applies it only where the worker owns preflight construction. DOCX, XLSX, worker-mode PPTX, progressive main mode, and Node paths were inspected and remain correctly scoped. No OOXML interpretation or heuristic is introduced; transfer, allocation, and lifecycle impact is negligible. Verified locally: focused worker/preflight/progressive tests (32), all PPTX tests (878), PPTX+core (4292 passed, 2 skipped), TypeScript build, ast-grep, and diff check. CI rust/audit/typecheck/smoke are green.

